PR #7, opened 2018-01-12
Replace redundant NULL check with assertion in bip32_key_init_alloc(...)

practicalswift:
Replace redundant `NULL` check with assertion in `bip32_key_init_alloc(...)`.


jgriffiths:
Thanks! merged in https://github.com/ElementsProject/libwally-core/pull/10

PR #9, opened 2018-01-12
Replace redundant NULL check with assertion in bip39_get_word(...)

practicalswift:
Replace redundant `NULL` check with assertion in `bip39_get_word(...)`.


jgriffiths:
Thanks! merged in https://github.com/ElementsProject/libwally-core/pull/10

PR #11, opened 2018-01-18
Remove bundled copy of libsecp256k1, and build using the system shared library

luke-jr:


jgriffiths:
Hi,

Making elements optional seems iike a good idea, but removing secp to rely on a system install isn't, since the system version may not have the config options we need, may not be present, doesn't work on windows, and makes it difficult/perhaps impossible to build the wrappers for python/node.

luke-jr:
Then the system can rebuild with those config options, and/or the user can install another copy for libwally...

Why wouldn't it work on Windows or build the wrappers?

Bundling libraries is a bad practice and discouraged, if not outright forbidden, by most major distros.

jgriffiths:
We can consider it when core uses the system secp perhaps, the same arguments apply there. However its already painful to build on all the supported platforms, and they don't come with a shared secp (in at least the Android case you are forced to bundle your library).

I could support a change that optionally uses the system secp without removing the code in tree. That would need to support disabling elements and having the swig wrappers not break too though, which will need more changes than in your second diff.

luke-jr:
No, it is part of consensus code in Core, so the same arguments do not apply.

Bundling the library in binaries is not made any more difficult by de-bundling the code. You can even build the lib static if you like, and it will just work...

(I need to figure out how to build the wrappers...)

greenaddress:
Luke,

We do not plan to ship any of the wallets that are based off libwally to distributions as otherwise they'd soon be old (for example see version of electrum

Having the option to link to a system library is OK but it has to be a build time option rather than the only option for us to support Android/iOS/AppImage.

AppImage is what we plan to use in the short term to ship wallets to linux distro without relying on distro release timings (but open to change that as the market picks a winner between appimage, snap, flatpack, etc)

Sjors:
> Having the option to link to a system library is OK but it has to be a build time option rather than the only option for us to support Android/iOS/AppImage.

More generally, IIUC, it has to be a build time option for any cross compilation to a sandboxed system.

Using a git submodule or subtree for `/src/secp256k1` would also be an improvement.

Sjors:
libwally now uses a submodule

jgriffiths:
Yes thanks for the remonder @Sjors , closing.

luke-jr:
Submodule is not comparable.

Sjors:
I agree it's different. Just means this PR would need a rebase.

I would also rather not have to rely on system distributions, if only because they don't exist for iOs :-)

PR #31, opened 2018-05-07
bip38: Use less confusing hash shuffling (BIP38_KEY_SWAP_ORDER)

practicalswift:
Use a more direct way to reference the last four bytes of `buf.half2`:

```
-        memcpy(buf.decode_hash - sizeof(uint32_t), &tmp, sizeof(uint32_t));
+        memcpy(buf.half2 + AES_BLOCK_LEN - sizeof(uint32_t), &tmp, sizeof(uint32_t));
```

The current code is being misunderstood by at least one static analyzer that is reporting a buffer overflow here due to the indirect way we're referencing `buf.half2`.


practicalswift:
@jgriffiths Would you be open to changes such as this one switching to more explicit copying on a per member level instead of copying multiple fields at once using knowledge about struct member ordering?

The current method cause some static analyzer false positives that I'd like to get rid of to increase the static analyzer signal-to-noise ratio :-)

PR #33, opened 2018-05-07
bip38: Make sure we're never operating on uninitialized variables

practicalswift:
Make sure we're never operating on uninitialized variables.

### DIFF
--- a/src/bip38.c
+++ b/src/bip38.c
@@ -222,7 +222,7 @@ static int to_private_key(const char *bip38,
                           unsigned char *bytes_out, size_t len)
 {
     struct derived_t derived;
-    struct bip38_layout_t buf;
+    struct bip38_layout_t buf = {};
     int ret = WALLY_EINVAL;
 
     if (flags & ~BIP38_ALL_DEFINED_FLAGS)


jgriffiths:
Hi @practicalswift,

Have you seen an actual uninitialized read of any part of this struct?

practicalswift:
@jgriffiths No, I haven't been able to trigger it myself but judging from the logic it seems to be possible to reach …

```
     if (buf.prefix != BIP38_PREFIX ||
         buf.flags & BIP38_FLAGS_RESERVED ||
…
```

… and then read an unitialized `buf.flags` given the right combination of `flags`, `bytes_out`, `bytes_len` and `len`?

I found this via a static analyzer.

jgriffiths:
Hi,

the flags are after the prefix in memory, and at least BIP38_SERIALIZED_LEN bytes must be written starting from prefix (either by mempcy or base58 decoding) before that code executes. So flags will always be written before reading in this case.

In general i'm against blanket initializing structs/vars before use since if the compiler doesn't remove the initialization as redundant, then tools like valgrind won't report invalid reads.

If your static analyser of choice has pragmas/comments etc to disregard false positives, im happy to take a patch that disables e.g. this one. Alternately you may want to file a bug with the tools authors.

practicalswift:
Thanks for clarifying! I'll investigate.

PR #36, opened 2018-06-11
Fix _FORTIFY_SOURCE redefined warnings

jsarenik:
The fix is the same as in libsodium, 4074470a
https://github.com/jedisct1/libsodium/commit/4074470a


jgriffiths:
Merged, thanks!

PR #50, opened 2018-09-04
Cast C++ output container sizes to size_t

jgriffiths:
This works around warnings from container wrappers such as some gsl implementations which return ptrdiff_t (a signed type) for size().


jkauffman1:
utAck 51709a91b228240e6c26207a382ba972ac701c22

greenaddress:
utack https://github.com/ElementsProject/libwally-core/commit/51709a91b228240e6c26207a382ba972ac701c22

PR #51, opened 2018-09-05
Add py3 testing on gitlab_ci

LeoComandini:
In addition, if py3, force utf-8 encoding when opening files containing wordlists.


greenaddress:
ack 359c01f783c24586a9ab465f1ca4620c0f8d3885

Issue #58, opened 2018-10-10
Rework struct to handle BIP38_KEY_SWAP_ORDER memory shuffling better


jgriffiths:
Both clangs static analyzer and some broken gcc version get tripped up by the memory shuffling this flag requires, as it causes memcpys across members within the same struct.

Both should be fixed by copying the data out to a temporary array, poking the values back in the correct order and securely nuking the temp buffer.

See https://github.com/ElementsProject/lightning/issues/1961 for an example of gcc b0rking on this.

jgriffiths:
fixed by #60

PR #59, opened 2018-10-10
Clean up BIP38_KEY_SWAP_ORDER handling to avoid compiler/analyzer bugs

jgriffiths:
For issue #58 - needs test cases


jgriffiths:
included in #60 with secp updates etc

PR #73, opened 2018-11-14
Enable BUILD_ELEMENTS flag in node-gyp compilation

dgyoshi:
`ENABLE_ELEMENTS=--enable-elements ./tools/build_js_bindings.sh` did not define `BUILD_ELEMENTS` flag. So it did not produce `wallycore.node` with elements features.

This makes Makefile to generate appropriate binding.gyp by --enable-elements option.


jgriffiths:
merged, thanks!

PR #80, opened 2019-01-14
Add bip39 functions to node.js

dgyoshi:
Add bip39_mnemonic_to_seed to node.js
Add bip39_mnemonic_from_bytes to node.js


greenaddress:
hi @dgyoshi and thank you for the PR!

I merged your commit here https://github.com/ElementsProject/libwally-core/commit/d97f67a74afdb3e69c64fe1652a6f394fe6c3b8c with some nits applied (rebased, removed a header that wasn't needed, slightly simplified the test, removed some white space and added the test to the makefile so that it runs with the tests when the JS bindings are enabled.

thank you again for the contribution and let me know if you have any comment and if we can close this PR.

PR #91, opened 2019-02-28
Add support for abbreviated bip39 mnemonic elements

### LINKED ISSUE #89: Support use of abbreviated 4-letter BIP39 words
The BIP39 wordlists are selected so that the first 4 letters of each word are unique. The functions for working with mnemonics should accept them in that form, and expand them to the full words.

This is important because tools like the CryptoSteel / BillFodl metal key storage devices only accept the first four letters of each word, and making the user hand-expand that back to the whole word is painful and unnecessary.

Screwing around with the input in this way playing guess-what-I-mean is fine because of the checksum. As long as the checksum passes, we know we performed the correct expansion. (In light of this I think we should also offer user-friendly things like automatic casefolding to the same case as the dictionary. All of this should only be triggered after trying the entered string as-is and failing the checksum in that form.)

mword:
Here's a simple set of changes that allow for the support of 4 (and greater) character bip39 mnemonic elements. It's tested and works.

One caveat with these changes is that it assumes that wordlist is only intended to handle 4 significant character words. If it's intended to be more general then the number of significant characters should be parameterized.

May resolve #89


greenaddress:
Rather than changing the current code, I think it would be best if all we did was adding a new function that, provided a 4 characters words, returns the matching full word - then users of wally can easily convert from the short form to the full form.

mword:
@greenaddress I can see a were a tool to convert short-form to full-form mnemonic strings might be useful but the two step process of 1. converting short-form to full-form string and then 2. doing full word string compares on the full words seems unnecessarily inefficient (as opposed to just 1. doing 4 character compares on any-form mnemonic. Maybe I'm missing something?

Also, with the changes I made, to get a full-form mnemonic string from a short-form string one need only do a `mnemonic_to_bytes` followed by a `mnemonic_from_bytes`. In fact, along with cloning wordlist_lookup_word (as wordlist_lookup_abbreviated_word?) I'd probably want to do just that to do the conversion.

Your call.

gwillen:
@greenaddress I have a counterproposal -- can we add a "unique_prefix" flag to wordlist_lookup_word, default false, which enables the following behavior: if there is exactly one word in the wordlist with "word" as a prefix, return that word, otherwise fail. (Alternatively this could be a new wordlist_lookup_word_unique_prefix function.) Then add the same flag to mnemonic_to_bytes and pass it through.

This doesn't require any special handling for unicode, doesn't require hardcoding the four-character prefix rule for bip39 wordlists (but works with any correct use of four-letter truncated mnemonics), and generally does exactly what we want with minimal changes.

PR #92, opened 2019-04-08
Add build target 'script.c, transaction.c' for build.bat

dgyoshi:
Script.c and transaction.c should be included in the build.bat.


greenaddress:
@dgyoshi thanks!

PR #98, opened 2019-05-09
fix issue 

HunterSun2008:
fix issue that the parameter 'bytes' of function WALLY_HMAC_FUNCTION cannot be set to NULL


greenaddress:
hi @HunterSun2008 what is the purpose of this change and what issue does it fix?

PR #99, opened 2019-05-15
Add functions to handle unique prefix mnemonics

gwillen:
Common practice with bip39 mnemonics is to use unique four-character prefixes
of wordlist words. In some cases (e.g. "steel" storage devices) there is not
space to store more than four characters. So, enable users to enter unique
prefixes instead of whole words.

This is an alternative to #91 for fixing #89 . If you like the approach, I can add tests and so forth. Unfortunately I forgot that C doesn't have default parameters, which makes this slightly messier than it would have been in C++. I also couldn't take advantage of bsearch due to the interface, which is slightly annoying but shouldn't be a big deal (wordlists are very short, time to search them should be negligible.)

### DIFF
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -83,6 +83,24 @@ size_t wordlist_lookup_word(const struct words *w, const char *word)
     return found ? found - w->indices + 1u : 0u;
 }
 
+size_t wordlist_lookup_prefix(const struct words *w, const char *prefix)
+{
+    const size_t prefix_len = strlen(prefix);
+    size_t found = 0;
+
+    size_t i;
+    for (i = 0; i < w->len; ++i) {
+        if (!strncmp(prefix, w->indices[i], prefix_len)) {
+            if (found) {
+                return 0;  // prefix match is not unique
+            }
+            found = i + 1;
+        }
+    }
+
+    return found;
+}
+
 const char *wordlist_lookup_index(const struct words *w, size_t idx)
 {
     if (idx >= w->len)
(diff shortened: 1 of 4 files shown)


gwillen:
(to avoid duplication, it might actually make sense to change the original interface instead of adding another one, if you like this approach -- but adding an interface is certainly more backwards-compatible. I think making this the main interface should be okay though. It would eliminate the use of bsearch, but as mentioned above, the amount of time this takes should be absolutely negligible, and it should never be getting called in a tight loop anyway.)

greenaddress:
concept ack, looking good to add tests IMHO

I think the new function approach is preferable here because of backwards compatibility - if it was something we had just added or that we were still iterating a bit on like liquid there'd be a stronger argument to break ABI but not so much in this case and not without leaving the prior function as deprecated but around for a while (which we can still do later if we decide that's the way forward).

src/wordlist.c
```diff
@@ -83,6 +83,24 @@ size_t wordlist_lookup_word(const struct words *w, const char *word)
     return found ? found - w->indices + 1u : 0u;
 }
 
+size_t wordlist_lookup_prefix(const struct words *w, const char *prefix)
```
instagibbs: this breaks with `act` versus `actual` mnemonic words, since `act` prefix matches for 2+ entries.

I think we should restrict the `prefix` input to length 3 or 4, and if length 3 it must be an exact match.

instagibbs:
suggested fix and test here: https://github.com/instagibbs/libwally-core/tree/prefix_fix

PR #100, opened 2019-05-28
Add MacOS artefacts to .gitignore

Sjors:
MacOS has the annoying habit of putting `.DSStore` files everywhere Finder goes.

I also found that `make` produced two `.sh-e` files, so I added a filter for that:
```
src/secp256k1/build-aux/ltmain.sh-e
tools/build-aux/ltmain.sh-e
```


greenaddress:
merged - thanks!

Issue #102, opened 2019-05-29
Derive address from BIP32 extended key


Sjors:
A good approach could be a variant on [wally_wif_to_address](https://wally.readthedocs.io/en/release_0.7.0/address/#c.wally_wif_from_bytes):

```c
int wally_addr_from_bip32_key(const struct ext_key *hdkey, const char *addr_family,\
                              uint32_t flags, char **output)
```

Where `flags` indicate the address type (legacy, native segwit, p2sh-segwit).

Most of the implementation can be borrowed from c-lightning, in particular [encode_pubkey_to_addr](https://github.com/ElementsProject/lightning/blob/11412435ce0cae92205178db9a9f4b64d0d78830/wallet/walletrpc.c#L200-L248) which uses a [pubkey straight from ext_key](https://github.com/ElementsProject/lightning/blob/master/wallet/walletrpc.c#L319-L336).

greenaddress:
I think this is implemented now, closing the issue, feel free to reopen otherwise.

PR #103, opened 2019-05-30
Derive P2PKH, P2WPKH  and P2SH-P2WPKH address from BIP32 key

Sjors:
Introduces function to derive a legacy (P2PKH) or wrapped SegWit (P2SH-P2WPKH) address from a BIP32 key:

```c
int wally_bip32_key_to_address(const struct ext_key *hdkey, uint32_t flags,
                               uint32_t version, char **output)
```

Implements #102 except for native SegWit, which I'll add a separate function for.

I did not add `c++` headers in `wally.hpp`.

I added tests in Python, but not for Javascript, Swig Python and Swig Java.

### DIFF
--- a/src/address.c
+++ b/src/address.c
@@ -54,3 +54,20 @@ int wally_bip32_key_to_address(const struct ext_key *hdkey, uint32_t flags,
     wally_clear(address, sizeof(address));
     return ret;
 }
+
+int wally_bip32_key_to_addr_segwit(const struct ext_key *hdkey, const char *addr_family,
+                                   uint32_t flags, char **output)
+{
+    int ret;
+
+    // Witness program bytes, including the version and data push opcode.
+    unsigned char witness_program_bytes[HASH160_LEN + 2];
+    witness_program_bytes[0] = OP_0;
+    witness_program_bytes[1] = HASH160_LEN;
+
+    if (wally_hash160(hdkey->pub_key, sizeof(hdkey->pub_key), witness_program_bytes + 2, HASH160_LEN) != WALLY_OK)
+        return WALLY_EINVAL;
+
+    ret = wally_addr_segwit_from_bytes(witness_program_bytes, HASH160_LEN + 2, addr_family, flags, output);
+    return ret;
+}
--- a/src/address.c
+++ b/src/address.c
@@ -22,10 +22,10 @@ int wally_bip32_key_to_address(const struct ext_key *hdkey, uint32_t flags,
         return WALLY_EINVAL;
 
     // Catch known incorrect combinations of address type and version:
-    if ((flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0x05) ||
-        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x00) ||
-        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0xC4) ||
-        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x6F))
+    if ((flags & WALLY_ADDRESS_TYPE_P2PKH && version == WALLY_ADDRESS_VERSION_P2SH_MAINNET) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == WALLY_ADDRESS_VERSION_P2PKH_MAINNET) ||
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == WALLY_ADDRESS_VERSION_P2SH_TESTNET) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == WALLY_ADDRESS_VERSION_P2PKH_TESTNET))
         return WALLY_EINVAL;
 
     if (flags == WALLY_ADDRESS_TYPE_P2PKH) {
--- a/src/address.c
+++ b/src/address.c
@@ -1,0 +1,56 @@
+#include "internal.h"
+#include "base58.h"
+#include "ccan/ccan/build_assert/build_assert.h"
+#include <include/wally_address.h>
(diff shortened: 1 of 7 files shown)


Sjors:
@glslang thanks for your review. Rebased and hopefully addresses all comments.

1. I ran `./tools/uncrustify`, but it only fixed one indentation. Is there another linter that you're running, or is this all manual? I renamed `pubKey(Hash)` to `pub_key(_hash)`, `redeemScript(Hash)` to `redeem_script(_hash)`. I removed braces around  `return WALLY_EINVAL`.

2. Fixed

3. Fixed (2x)

glslang:
unfortunately is manual currently. many thanks for the changes.

Sjors:
Added native SegWit in a separate commit.

src/address.c
```diff
@@ -0,0 +1,81 @@
...
+
+    if (flags == WALLY_ADDRESS_TYPE_P2PKH) {
+        // pub_key_hash = ripemd160(sha256(pubkey)
+        address[0] = (unsigned char) version & 0xff;
+        if (wally_hash160(hdkey->pub_key, sizeof(hdkey->pub_key), address + 1, HASH160_LEN) != WALLY_OK)
+            return WALLY_EINVAL;
+
+        ret = wally_base58_from_bytes(address, sizeof(address), BASE58_FLAG_CHECKSUM, output);
+    } else {
+        BUILD_ASSERT(flags == WALLY_ADDRESS_TYPE_P2SH_P2WPKH);
```
jkauffman1: Should this be a runtime check?

src/address.c
```diff
@@ -0,0 +1,81 @@
...
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x6F)
+        ) {
+        return WALLY_EINVAL;
+    }
+
+    unsigned char address[HASH160_LEN + 1];
+
+    if (flags == WALLY_ADDRESS_TYPE_P2PKH) {
+        // pub_key_hash = ripemd160(sha256(pubkey)
+        address[0] = (unsigned char) version & 0xff;
```
jkauffman1: Could possibly be pulled out of the if

src/address.c
Sjors: This should never happen, because of the above `Catch known incorrect combinations` checks. I don't know if the compiler is smart enough to understand that though. If not, then I should make it a runtime check (but I couldn't get `assert` to work).

Sjors:
Addressed @jkauffman1's suggestion to move `ret = wally_base58_from_bytes(...` out of the if statement. Not sure what to do about the `BUILD_ASSERT`.

Added testnet vectors.

src/address.c
glslang: the comment has more to do with the fact that this isn't a compile time expression and will cause a sizeof evaluatiion of a VLA effectively. I think it needs to be converted to an if expression or removed.

Sjors: This goes way beyond my C comprehension. I'll just drop the `BUILD_ASSERT` then.

Sjors:
Travis failure looks random to me (`No available formula with the name "swig@3.04"`)

src/address.c
```diff
@@ -0,0 +1,77 @@
...
+    if (
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0x05) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x00) ||
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0xC4) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x6F)
+        ) {
+        return WALLY_EINVAL;
+    }
+
+    unsigned char address[HASH160_LEN + 1];
```
glslang: style: move to before "int ret" on line 12

src/address.c
```diff
@@ -0,0 +1,77 @@
...
+    if (!(flags & (WALLY_ADDRESS_TYPE_P2PKH | WALLY_ADDRESS_TYPE_P2SH_P2WPKH)))
+        return WALLY_EINVAL;
+
+    // Catch known incorrect combinations of address type and version:
+    if (
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0x05) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x00) ||
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0xC4) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x6F)
+        ) {
```
glslang: nit: remove { to conform to code style

Sjors:
Done and rebased.

src/address.c
```diff
@@ -0,0 +1,73 @@
...
+        return WALLY_EINVAL;
+
+    if (!(flags & (WALLY_ADDRESS_TYPE_P2PKH | WALLY_ADDRESS_TYPE_P2SH_P2WPKH)))
+        return WALLY_EINVAL;
+
+    // Catch known incorrect combinations of address type and version:
+    if ((flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0x05) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x00) ||
+        (flags & WALLY_ADDRESS_TYPE_P2PKH && version == 0xC4) ||
+        (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == 0x6F))
```
jkauffman1: Maybe it's worth defining these constants (0, 5, 111, 196) especially as they are being used in the related PR: https://github.com/ElementsProject/libwally-core/pull/112

Sjors: Done

greenaddress:
utack eb5257cb314dab9e9fe17bb360cdf03b88f17e97

PR #112, opened 2019-06-11
Parse address to scriptPubKey

Sjors:
My initial idea was to write a parser that can take either a base58 or bech32 address and return a `scriptPubKey`, with flags to constrain it to correct coin. However, to keep the change minimal, I'm using a approach that reuses `wally_addr_segwit_to_bytes` for bech32 and adds a new function for base58. A caller should either figure out themselves which address format they're dealing with, or simply try both functions. 

This PR adds `wally_address_to_scriptpubkey` to parse a base58 addresses (P2PKH and P2SH). It returns a `scriptPubKey` which can be used to create a transaction output.

A flag is used to set either Bitcoin mainnet or testnet. It then automatically figures out if this is a `P2PKH` or `P2SH` address. This approach requires additional `switch` cases to support other coins.

An alternative approach could be to add a `version` argument. The downside of that approach is that the caller needs to loop over all possible address versions they support, e.g. call with `0x00` and `0x05` to try if a given address is mainnet P2PKH or P2SH. Although that approach is more similar to other libwally functions that take a version argument, I don't find it very user friendly.

This PR also adds a test for the existing `wally_addr_segwit_to_bytes` which parses bech32 addresses and returns a `scriptPubKey`.

This PR builds on top of #102, so the first 3 commits should be reviewed there.

### DIFF
--- a/include/wally_address.h
+++ b/include/wally_address.h
@@ -15,6 +15,9 @@ struct ext_key;
 #define WALLY_CA_PREFIX_LIQUID 0x0c /** Liquid v1 confidential address prefix */
 #define WALLY_CA_PREFIX_LIQUID_REGTEST 0x04 /** Liquid v1 confidential address prefix for regtest */
 
+#define WALLY_NETWORK_BITCOIN_MAINNET 0x01 /** Bitcoin mainnet */
+#define WALLY_NETWORK_BITCOIN_TESTNET 0x02 /** Bitcoin testnet */
+
 #define WALLY_ADDRESS_TYPE_P2PKH 0x01       /** P2PKH address ("1...") */
 #define WALLY_ADDRESS_TYPE_P2SH_P2WPKH 0x02 /** P2SH-P2WPKH wrapped SegWit address ("3...") */
 #define WALLY_ADDRESS_TYPE_P2WPKH 0x04      /** P2WPKH native SegWit address ("bc1...)" */
@@ -53,6 +56,22 @@ WALLY_CORE_API int wally_addr_segwit_from_bytes(
 WALLY_CORE_API int wally_addr_segwit_to_bytes(
     const char *addr,
     const char *addr_family,
+    uint32_t flags,
+    unsigned char *bytes_out,
+    size_t len,
+    size_t *written);
+
+/**
+ * Infer a scriptPubKey from an address.
+ *
+ * :param addr: Address to infer the scriptPubKey from.
+ * :param flags: Pass ``WALLY_NETWORK_BITCOIN_MAINNET`` or ``WALLY_NETWORK_BITCOIN_TESTNET``.
+ * :param bytes_out: Destination for the resulting scriptPubKey
+ * :param len: Length of ``bytes_out`` in bytes.
+ * :param written: Destination for the number of bytes written to ``bytes_out``.
+ */
+WALLY_CORE_API int wally_address_to_scriptpubkey(
+    const char *addr,
     uint32_t flags,
     unsigned char *bytes_out,
     size_t len,
--- a/src/address.c
+++ b/src/address.c
@@ -1,5 +1,6 @@
 #include "internal.h"
 #include "base58.h"
+#include <stdbool.h>
 #include "ccan/ccan/build_assert/build_assert.h"
 #include <include/wally_address.h>
 #include <include/wally_bip32.h>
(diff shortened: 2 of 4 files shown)


src/address.c
```diff
@@ -0,0 +1,145 @@
...
+    // Check minimum length of bytes_out:
+    // P2PKH 25 bytes: OP_DUP OP_HASH160 [pub_key_hash] OP_EQUALVERIFY OP_CHECKSIG
+    // P2SH  23 bytes: OP_HASH160 [script_hash] OP_EQUAL
+    if (len < 25)
+        return WALLY_EINVAL;
+
+    if (wally_base58_to_bytes(addr, BASE58_FLAG_CHECKSUM, bytes_base58_decode, sizeof(bytes_base58_decode), &written_base58_decode) != WALLY_OK)
+        return WALLY_EINVAL;
+
+    if (written_base58_decode == 0)
```
glslang: don't think the call above can ever return 0 as written and WALLY_OK

src/address.c
```diff
@@ -0,0 +1,145 @@
...
+    return ret;
+}
+
+int wally_address_to_scriptpubkey(const char *addr, uint32_t flags, unsigned char *bytes_out,
+                                  size_t len, size_t *written)
+{
+    int ret;
+
+    uint32_t version;
+    unsigned char bytes_base58_decode[len + 1];
```
glslang: this forces a VLA which we don't want. there are examples elsewhere for this conversion and how to declare the array. see elements.c for example.

src/address.c
Sjors: I'll double check `wally_base58_to_bytes` and drop the `if`.

src/address.c
Sjors: I'll look at the elements code to see how to avoid a variable length array. I guess I can just use the worst of `len`.

include/wally_address.h
```diff
@@ -47,6 +56,22 @@ WALLY_CORE_API int wally_addr_segwit_to_bytes(
     size_t len,
     size_t *written);
 
+/**
+ * Infer a scriptPubKey from an address.
+ *
+ * :param addr: Address to infer the scriptPubKey from.
+ * :param flags: Pass ``WALLY_NETWORK_BITCOIN_MAINNET`` or ``WALLY_NETWORK_BITCOIN_TESTNET``.
```
jkauffman1: nit: I'm not sure I would call these "flags" as they are mutually exclusive

src/address.c
```diff
@@ -0,0 +1,145 @@
...
+    size_t written_base58_decode;
+
+    if (written)
+        *written = 0;
+
+    if ((flags & ~0xff))
+        return WALLY_EINVAL;
+
+    if (!(flags & (WALLY_NETWORK_BITCOIN_MAINNET | WALLY_NETWORK_BITCOIN_TESTNET)))
+        return WALLY_EINVAL;
```
jkauffman1: It may be simpler to just check that "flags" is equal to either WALLY_NETWORK_BITCOIN_MAINNET or WALLY_NETWORK_BITCOIN_TESTNET and then drop the check on line 87/88 as well.

src/address.c
```diff
@@ -0,0 +1,145 @@
...
+
+    // Check minimum length of bytes_out:
+    // P2PKH 25 bytes: OP_DUP OP_HASH160 [pub_key_hash] OP_EQUALVERIFY OP_CHECKSIG
+    // P2SH  23 bytes: OP_HASH160 [script_hash] OP_EQUAL
+    if (len < 25)
+        return WALLY_EINVAL;
+
+    if (wally_base58_to_bytes(addr, BASE58_FLAG_CHECKSUM, bytes_base58_decode, sizeof(bytes_base58_decode), &written_base58_decode) != WALLY_OK)
+        return WALLY_EINVAL;
```
jkauffman1: It seems wally_base58_to_bytes can return WALLY_OK to indicate that the output buffer was not big enough. May need to explicitly check for that condition.

src/address.c
```diff
@@ -0,0 +1,145 @@
...
+        if (flags & WALLY_NETWORK_BITCOIN_TESTNET)
+            return WALLY_EINVAL;
+    case 111: // Bitcoin testnet P2PKH
+        if (version == 111 && flags & WALLY_NETWORK_BITCOIN_MAINNET)
+            return WALLY_EINVAL;
+        *written = HASH160_LEN + 5;
+        bytes_out[0] = OP_DUP;
+        bytes_out[1] = OP_HASH160;
+        bytes_out[2] = HASH160_LEN;
+        memcpy(bytes_out + 3, bytes_base58_decode + 1, *written);
```
jkauffman1: Should this be copying HASH160_LEN bytes (rather than HASH160_LEN + 5)?

Sjors: Yes, thanks!

src/address.c
Sjors: This indeed can't happen, because `if (*written <= BASE58_CHECKSUM_LEN) {` would have returned `WALLY_EINVAL`. But having the check here anyway seems more robust to me; perhaps later someone modifies `wally_base58_to_bytes` to handle an empty string more gracefully?

src/address.c
Sjors: I added a check that `wally_base58_to_bytes` returns `written` exactly `HASH160_LEN + 1`.

include/wally_address.h
Sjors: We use mutually exclusive flags for `EC_FLAG_ECDSA` and `EC_FLAG_SCHNORR` too.

src/address.c
Sjors: Done

src/address.c
Sjors: I just made it `HASH160_LEN + 1`

src/address.c
```diff
@@ -0,0 +1,149 @@
...
+    version = bytes_base58_decode[0];
+
+    switch (version) {
+    case 0: // Bitcoin P2PKH
+        if (flags & WALLY_NETWORK_BITCOIN_TESTNET)
+            return WALLY_EINVAL;
+    case 111: // Bitcoin testnet P2PKH
+        if (version == 111 && flags & WALLY_NETWORK_BITCOIN_MAINNET)
+            return WALLY_EINVAL;
+        if (written_base58_decode != HASH160_LEN + 1)
```
jkauffman1: Is there a reason why this check and the identical one on line 133 could not be moved up to line 102, and replace the check for == 0?

Sjors: I reorganized that code a little bit. It's actually longer, but probably more readable and less repetitive.
1. moved the input length check to a separate switch statement
2. moved the network check to a separate switch statement

src/address.c
```diff
@@ -0,0 +1,161 @@
...
+    switch (version) {
+    case 0: // Bitcoin P2PKH
+    case 111: // Bitcoin testnet P2PKH
+        *written = HASH160_LEN + 5;
+        bytes_out[0] = OP_DUP;
+        bytes_out[1] = OP_HASH160;
+        bytes_out[2] = HASH160_LEN;
+        memcpy(bytes_out + 3, bytes_base58_decode + 1, HASH160_LEN);
+        bytes_out[HASH160_LEN + 3] = OP_EQUALVERIFY;
+        bytes_out[HASH160_LEN + 4] = OP_CHECKSIG;
```
jkauffman1: We could possibly call wally_scriptpubkey_p2pkh_from_bytes here

src/address.c
```diff
@@ -0,0 +1,161 @@
...
+        ret = WALLY_OK;
+        break;
+
+     case 5: // Bitcoin P2SH
+     case 196: // Bitcoin testnet P2SH
+        *written = HASH160_LEN + 3;
+        bytes_out[0] = OP_HASH160;
+        bytes_out[1] = HASH160_LEN;
+        memcpy(bytes_out + 2, bytes_base58_decode + 1, HASH160_LEN);
+        bytes_out[HASH160_LEN + 2] = OP_EQUAL;
```
jkauffman1: And equally here we could call wally_scriptpubkey_p2sh_from_bytes

glslang:
equivalent gdk code here so we could expect this function to be equally concise,
https://github.com/Blockstream/gdk/blob/master/src/transaction_utils.cpp#L83

src/address.c
Sjors: Done

Sjors:
Switched to using `wally_scriptpubkey_p2pkh_from_bytes` and `wally_scriptpubkey_p2sh_from_bytes `.

src/address.c
```diff
@@ -0,0 +1,146 @@
...
+        *written = 0;
+
+    if (!(flags == WALLY_NETWORK_BITCOIN_MAINNET || flags == WALLY_NETWORK_BITCOIN_TESTNET))
+        return WALLY_EINVAL;
+
+    // Check minimum length of bytes_out:
+    // P2PKH 25 bytes: OP_DUP OP_HASH160 [pub_key_hash] OP_EQUALVERIFY OP_CHECKSIG
+    // P2SH  23 bytes: OP_HASH160 [script_hash] OP_EQUAL
+    if (len < 25)
+        return WALLY_EINVAL;
```
jkauffman1: I think now we are calling wally_scriptpubkey_p2pkh_from_bytes/wally_scriptpubkey_p2sh_from_bytes we can drop this explicit length check because those functions themselves check len and return WALLY_EINVAL if len is too small for their respective outputs.

Sjors: Indeed, those functions alraedy check `len < WALLY_SCRIPTPUBKEY_P2PKH_LEN` and `_P2SH_LEN`.

Sjors:
Rebased. Using `WALLY_ADDRESS_VERSION_` and dropped redundant length check.

Sjors:
Rebased now that #103 has landed.

Sjors:
@glslang better?

glslang:
utAck 4d6382f965ac27c8dc8a50bf562e6187c47ed94c

glslang:
> @glslang better?

I think this version is more inline with the rest of the code base. many thanks for the change.

PR #113, opened 2019-06-11
Clarify that wally_addr_segwit_from_bytes returns scriptPubKey

Sjors:
It returns more than just the witness program.


greenaddress:
utack e677e624377fda05c8842cfb63bb1675f8dae435

PR #115, opened 2019-06-14
support empty(0-byte) witness stack item in clone_witness().

k-matsuzawa:
The clone_witness function should support 0 byte witness item.

Currently, wally_tx_witness_stack_set() is supported 0 byte witness item.
However, due to a null check in clone_witness(), wally_tx_set_input_witness() don't support 0 byte witness items.
(If it is add 0 byte witness item to witness stack on wally_tx_witness_stack_add(), and it's add witness stack to txin on wally_tx_set_input_witness(), 0 byte witness item disappears from the witness stack on txin.)

This fix removes unnecessary checks of the witness items from clone_witness().
The witness items to be set should be checked with wally_tx_witness_stack_set().

(Please excuse my poor English.)

### DIFF
--- a/src/transaction.c
+++ b/src/transaction.c
@@ -216,13 +216,11 @@ static struct wally_tx_witness_stack *clone_witness(
 
     if (ret == WALLY_OK) {
         for (i = 0; i < stack->num_items && ret == WALLY_OK; ++i) {
-            if (stack->items[i].witness) {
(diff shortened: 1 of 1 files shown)


greenaddress:
@k-matsuzawa merged in b011eb469c64121e39e44590127eed543a71e187 - thanks!

PR #122, opened 2019-07-01
Don't clear null pointer

jkauffman1:
If the malloc on line 99 fails and returns a null pointer it should not
be passed to wally_clear.


greenaddress:
merged in 721b67c061b38bc5ec2c8e5cd41e89357701685d

PR #126, opened 2019-08-23
Implement BIP 174 PSBT Serialization

### LINKED ISSUE #53: Support BIP 174 PSBT
It would be nice for libwally to support BIP 174 PSBTs.

achow101:
Implements the data structures for PSBT, serialization and deserialization of PSBTs, and functions for updating and combining PSBTs.

The base64 module from ccan is added to allow for handling base64 encoded PSBTs as specified in the BIP.

Closes #53


src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+        !memcpy(&keypaths->items[latest].origin.fingerprint, fingerprint, 4)) {
+        return WALLY_ENOMEM;
+    }
+    if (path) {
+        if (!clone_bytes((unsigned char **)&keypaths->items[latest].origin.path, (unsigned char *)path, path_len * sizeof(*path))) {
+            return WALLY_ENOMEM;
+        }
+        keypaths->items[latest].origin.path_len = path_len;
+    }
+    keypaths->num_items++;
```
glslang: it'd appear if path is null it will still increase num_items so it seems to assume null paths can be added. is this the intention?

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+    result->num_items = 0;
+    return WALLY_OK;
+}
+
+int partial_sigs_map_free(struct partial_sigs_map *sigs)
+{
+    size_t i;
+
+    if (sigs) {
+        if (sigs->items) {
```
glslang: the two if statements could be merged in a single one

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+        (unknowns && !(input->unknowns = clone_unknowns_map(unknowns))))
+        ret = WALLY_ENOMEM;
+    if (ret == WALLY_OK) {
+        input->redeem_script_len = redeem_script_len;
+        input->witness_script_len = witness_script_len;
+        input->final_script_sig_len = final_script_sig_len;
+        input->sighash_type = sighash_type;
+    }
+
+    return ret;
```
glslang: this function can potentially leak memory.

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+        for (i = 0; i < psbt->num_outputs; ++i) {
+            wally_psbt_output_free(&psbt->outputs[i]);
+        }
+        wally_free(psbt->outputs);
+    }
+    psbt->inputs = wally_malloc(tx->num_inputs * sizeof(struct wally_psbt_input));
+    psbt->outputs = wally_malloc(tx->num_outputs * sizeof(struct wally_psbt_output));
+    if (!psbt->inputs || !psbt->outputs) {
+        return WALLY_ENOMEM;
+    }
```
glslang: possible memory leak

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+    // Allocate the decoded buffer
+    safe_len = base64_decoded_length(strlen(string));
+    if ((decoded = wally_malloc(safe_len)) == NULL) {
+        return WALLY_ENOMEM;
+    }
+
+    // Decode the base64 psbt
+    decoded_len = base64_decode(decoded, safe_len, string, strlen(string));
+    if (decoded_len <= 5) { // Make sure we also have enough bytes for the magic
+        return WALLY_EINVAL;
```
glslang: possible memory leak

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+        return WALLY_ENOMEM;
+    }
+
+    // Get psbt bytes
+    if ((ret = wally_psbt_to_bytes(psbt, buff, len, &written)) != WALLY_OK) {
+        goto done;
+    }
+
+    // Base64 encode
+    b64_safe_len = base64_encoded_le
```
glslang: appears to assume psbts is never null.

glslang: possible null dereference

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+    struct keypath_map *result;
+    size_t i;
+    int ret;
+
+    ret = keypath_map_init_alloc(keypaths->items_allocation_len, &result);
+
+    if (ret == WALLY_OK) {
+        for (i = 0; i < keypaths->num_items && ret == WALLY_OK; ++i) {
+            if (!memcpy(&result->items[i].pubkey, keypaths->items[i].pubkey, 65) ||
+                !memcpy(&result->items[i].origin.fingerprint, keypaths->items[i].origin.fingerprint, 4)) {
```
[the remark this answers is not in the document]
achow101: It looks like `memcpy_s` was introduced in C11, I'm not sure what C version is targeted here. At the very least, I don't see `memcpy_s` used elsewhere in libwally.

src/psbt.c
achow101: That is intended.

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+
+static struct keypath_map *clone_keypath_map(const struct keypath_map *keypaths)
+{
+    struct keypath_map *result;
+    size_t i;
+    int ret;
+
+    ret = keypath_map_init_alloc(keypaths->items_allocation_len, &result);
+
+    if (ret == WALLY_OK) {
```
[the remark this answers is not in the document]
achow101: Done here and in the other `clone_*_map` functions.

src/psbt.c
```diff
@@ -0,0 +1,2302 @@
...
+                    ret = WALLY_ENOMEM;
+                }
+                result->items[i].origin.path_len = keypaths->items[i].origin.path_len;
+            }
+
+            if (ret != WALLY_OK) {
+                keypath_map_free(result);
+            }
+        }
+        result->num_items = keypaths->num_items;
```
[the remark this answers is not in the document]
achow101: Added early returns here and in the other `clone_*_map` functions.

src/psbt.c
achow101: Done here and in the other `clone_*_map` functions.

src/psbt.c
achow101: Fixed

src/psbt.c
achow101: Fixed

src/psbt.c
achow101: Added a check earlier to ensure this assumption is always true

achow101: How so?

src/psbt.c
achow101: I believe I've fixed that. Also fixed `wally_psbt_output_update` which had the same potential leaks.

src/psbt.c
```diff
@@ -0,0 +1,2387 @@
...
+    result->items_allocation_len = alloc_len;
+    result->num_items = 0;
+    return WALLY_OK;
+}
+
+int keypath_map_free(struct keypath_map *keypaths)
+{
+    size_t i;
+
+    if (keypaths && keypaths->items) {
```
jkauffman1: Might it be better to make clearing/freeing keypaths only dependent on keypaths. It looks like if you call keypath_map_init_alloc with alloc_len=0 it will allocate a keypath_map with no items which then won't be freed by keypath_map_free.

src/psbt.c
```diff
@@ -0,0 +1,2387 @@
...
+        if (keypaths->items[i].origin.path) {
+            if (!clone_bytes((unsigned char **)&result->items[i].origin.path, (unsigned char *)keypaths->items[i].origin.path, keypaths->items[i].origin.path_len * sizeof(*keypaths->items[i].origin.path))) {
+                goto fail;
+            }
+            result->items[i].origin.path_len = keypaths->items[i].origin.path_len;
+        }
+    }
+    result->num_items = keypaths->num_items;
+
+    return WALLY_OK;
```
jkauffman1: return result?

src/psbt.c
```diff
@@ -0,0 +1,2387 @@
...
+    size_t latest;
+
+    if (keypaths->num_items == keypaths->items_allocation_len) {
+        if (keypaths->items_allocation_len == 0) {
+            keypaths->items_allocation_len = 1;
+        }
+        keypaths->items_allocation_len *= 2;
+        struct keypath_map *new_keypaths = clone_keypath_map(keypaths);
+        if (!new_keypaths) {
+            return WALLY_ENOMEM;
```
jkauffman1: This code path looks like it leaves keypaths in an inconsistent state where items_allocation_len has been incremented, but the allocated buffer is still the original.

src/psbt.c
achow101: I removed the check for items here and in the other map free functions.

src/psbt.c
achow101: Fixed

src/psbt.c
achow101: Changed it to store the original allocation length and restore that if cloning fails.

src/psbt.c
glslang: there's no explicit check for a NULL psbts. get_psbt_txid doesn't check for NULL either.

include/wally_psbt.h
```diff
@@ -0,0 +1,454 @@
...
+ * :param non_witness_utxo: The non witness utxo for this input if it exists.
+ * :param witness_utxo: The witness utxo for this input if it exists.
+ * :param redeem_script: The redeem script for this input
+ * :param redeem_script_len: The length of the redeem script.
+ * :param witness_script: The witness script for this input
+ * :param witness_script_len: The length of the witness script.
+ * :param final_script_sig: The scriptSig for this input
+ * :param final_script_sig_len: Size of ``final_script_sig`` in bytes.
+ * :param final_witness: The witness stack for the input, or NULL if no witness is present.
+ * :param keypaths: The HD keypaths for ths input.
```
jkauffman1: typo: s/ths input/this input/

include/wally_psbt.h
```diff
@@ -0,0 +1,454 @@
...
+ * :param partial_sigs: The partial signatures for ths input.
```
jkauffman1: typo: s/ths input/this input/

include/wally_psbt.h
```diff
@@ -0,0 +1,454 @@
...
+ * :param unknowns: The unknown key value pairs for ths input.
```
jkauffman1: typo: s/ths input/this input/

include/wally_psbt.h
```diff
@@ -0,0 +1,454 @@
...
+
+/**
+ * Allocate and initialize a new psbt output.
+ *
+ * :param redeem_script: The redeem script needed for spending this output
+ * :param redeem_script_len: The length of the redeem script.
+ * :param witness_script: The witness script needed for spending for this output
+ * :param witness_script_len: The length of the witness script.
+ * :param keypaths: The HD keypaths for the keys needed for spending this output
+ * :param unknowns: The unknown key value pairs for ths output.
```
jkauffman1: typo: s/ths input/this input/

src/psbt.c
```diff
@@ -0,0 +1,2547 @@
...
+}
+
+int keypath_map_free(struct keypath_map *keypaths)
+{
+    size_t i;
+
+    if (keypaths) {
+        for (i = 0; i < keypaths->num_items; ++i) {
+            if (keypaths->items[i].origin.path) {
+                wally_clear(keypaths->items[i].origin.path, keypaths->items[i].origin.path_len);
```
jkauffman1: do the paths not need to be freed as well?

src/psbt.c
```diff
@@ -0,0 +1,2547 @@
...
+{
+    size_t latest;
+
+    if (keypaths->num_items == keypaths->items_allocation_len) {
+        size_t orig_alloc_len = keypaths->items_allocation_len;
+        if (keypaths->items_allocation_len == 0) {
+            keypaths->items_allocation_len = 1;
+        }
+        keypaths->items_allocation_len *= 2;
+        struct keypath_map *new_keypaths = clone_keypath_map(keypaths);
```
jkauffman1: This feels to me like an unnecessary memory allocation, couldn't it just allocate a new items array without allocating a new keypath_map?

src/psbt.c
```diff
@@ -0,0 +1,2547 @@
...
+static int clone_tx(struct wally_tx *tx, struct wally_tx **output)
+{
+    unsigned char *bytes;
+    size_t calc, written;
+    int ret = WALLY_OK;
+
+    TX_CHECK_OUTPUT;
+
+    if ((ret = wally_tx_get_length(tx, WALLY_TX_FLAG_USE_WITNESS, &calc)) != WALLY_OK) {
+        goto done;
```
jkauffman1: bytes is uninitialized here

src/psbt.c
```diff
@@ -0,0 +1,2547 @@
...
+    struct unknowns_map *unknowns)
+{
+    if (redeem_script) {
+        if (output->redeem_script) {
+            wally_free(output->redeem_script);
+        }
+        if (clone_bytes(&output->redeem_script, redeem_script, redeem_script_len)) {
+            output->redeem_script_len = redeem_script_len;
+        } else {
+            return WALLY_ENOMEM;
```
jkauffman1: In this code path output->redeem_script has been freed but other data members have not and the caller cannot tell which. Maybe set the freed pointers to null?

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+    wally_bzero(psbt->inputs, psbt->inputs_allocation_len * sizeof(struct wally_psbt_input));
+    psbt->num_inputs = tx->num_inputs;
+
+    for (i = 0; i < psbt->num_outputs; ++i) {
+        wally_psbt_output_free(&psbt->outputs[i]);
+    }
+    if (psbt->outputs_allocation_len < tx->num_outputs) {
+        wally_free(psbt->outputs);
+        psbt->outputs_allocation_len = 0;
+        psbt->outputs = wally_malloc(tx->num_outputs * sizeof(struct wally_psbt_input));
```
glslang: struct wally_psbt_output presumably

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+        // Process based on type
+        switch (type) {
+        case WALLY_PSBT_GLOBAL_UNSIGNED_TX: {
+            bool expect_wit;
+            if (analyze_tx(p, value_len, 0, &result->num_inputs, &result->num_outputs, &expect_wit) != WALLY_OK) {
+                return WALLY_EINVAL;
+            }
+            if ((result->input_counts = wally_malloc(result->num_inputs * sizeof(struct psbt_input_counts))) == NULL ||
+                (result->output_counts = wally_malloc(result->num_outputs * sizeof(struct psbt_output_counts))) == NULL) {
+                return WALLY_ENOMEM;
```
glslang: possible leak of result->input_counts

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+        }
+        case WALLY_PSBT_IN_PARTIAL_SIG: {
+            struct partial_sigs_map *partial_sigs = result->partial_sigs;
+            if (key_len != 66 && key_len != 34) {
+                return WALLY_EINVAL;     // Size of key is unexpected
+            }
+            // Check for duplicates
+            for (i = 0; i < partial_sigs->num_items; ++i) {
+                if (memcmp(partial_sigs->items[i].pubkey, &key[1], key_len - 1) == 0) {
+                    return WALLY_EINVAL;     // Duplicate key
```
glslang: possible leaks of all the memory allocated above it

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+    // Magic
+    memcpy(p, WALLY_PSBT_MAGIC, 5);
+    p += 5;
+
+    // Global tx
+    type = WALLY_PSBT_GLOBAL_UNSIGNED_TX;
+    p += varbuff_to_bytes(&type, 1, p);
+    ret = wally_tx_get_length(psbt->tx, 0, &tx_len);
+    if (ret != WALLY_OK) {
+
```
glslang: buff is always non null if it reaches here

glslang: should check for NULL str above. also cache strlen since is called twice.

include/wally_psbt.h
```diff
@@ -0,0 +1,430 @@
...
+ * Add an item to a keypath map
+ *
+ * :param keypaths: The keypath map to add to
+ * :param pubkey: The pubkey to add
+ * :param fingerprint: The master key fingerprint for the pubkey
+ * :param path: The BIP32 derivation path for the pubkey
+ * :param path_len: The number of items in path
+ */
+WALLY_CORE_API int add_new_keypath(struct keypath_map *keypaths,
+                                   unsigned char pubkey[65],
```
jkauffman1: Could these instances of literal 65 be EC_PUBLIC_KEY_UNCOMPRESSED_LEN?

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+static bool clone_bytes(unsigned char **dst, const unsigned char *src, size_t len)
+{
+    if (!len)
+        return true;
+    *dst = wally_malloc(len);
+    if (*dst)
+        memcpy(*dst, src, len);
+    return *dst != NULL;
+}
```
jkauffman1: Would it be possible to move this function into maybe internal.c and reuse it from psbt.c and transaction.c?

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+
+    for (i = 0; i < psbt->num_inputs; ++i) {
+        wally_psbt_input_free(&psbt->inputs[i]);
+    }
+    if (psbt->inputs_allocation_len < tx->num_inputs) {
+        wally_free(psbt->inputs);
+        psbt->inputs_allocation_len = 0;
+        psbt->inputs = wally_malloc(tx->num_inputs * sizeof(struct wally_psbt_input));
+        if (!psbt->inputs) {
+            return WALLY_ENOMEM;
```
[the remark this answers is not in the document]
achow101: How so?

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+    for (i = 0; i < psbt->num_outputs; ++i) {
+        wally_psbt_output_free(&psbt->outputs[i]);
+    }
+    if (psbt->outputs_allocation_len < tx->num_outputs) {
+        wally_free(psbt->outputs);
+        psbt->outputs_allocation_len = 0;
+        psbt->outputs = wally_malloc(tx->num_outputs * sizeof(struct wally_psbt_input));
+        psbt->outputs_allocation_len = tx->num_outputs;
+        if (!psbt->outputs) {
+            return WALLY_ENOMEM;
```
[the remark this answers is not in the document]
achow101: How so?

src/psbt.c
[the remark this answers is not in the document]
glslang: psbt->tx is allocated and is not freed here if this fails. otherwise leaves psbt with some side effect

src/psbt.c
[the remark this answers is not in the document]
glslang: same as above or leaves them allocated as a side effect.

src/psbt.c
achow101: Should I also move other things that are shared like `TX_CHECK_OUTPUT`, `analyze_tx`, etc or should those go in a more transaction specific internal thing?

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+        keypaths->num_items = orig_num_items;
+        keypaths->items_allocation_len = new_alloc_len;
+    }
+
+    latest = keypaths->num_items;
+
+    if (!memcpy(&keypaths->items[latest].pubkey, pubkey, 65) ||
+        !memcpy(&keypaths->items[latest].origin.fingerprint, fingerprint, 4)) {
+        return WALLY_ENOMEM;
+    }
```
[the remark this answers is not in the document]
achow101: Done

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+}
+
+int keypath_map_free(struct keypath_map *keypaths)
+{
+    size_t i;
+
+    if (keypaths) {
+        for (i = 0; i < keypaths->num_items; ++i) {
+            if (keypaths->items[i].origin.path) {
+                clear_and_free(keypaths->items[i].origin.path, keypaths->items[i].origin.path_len);
```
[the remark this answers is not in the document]
achow101: fixed

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+        sigs->items = new_items;
+        sigs->num_items = orig_num_items;
+        sigs->items_allocation_len = new_alloc_len;
+    }
+
+    latest = sigs->num_items;
+
+    if (!memcpy(&sigs->items[latest].pubkey, pubkey, 65)) {
+        return WALLY_ENOMEM;;
+    }
```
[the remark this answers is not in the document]
achow101: Done

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+}
+
+int wally_psbt_set_global_tx(
+    struct wally_tx *tx,
+    struct wally_psbt *psbt)
+{
+    int ret = WALLY_OK;
+    size_t i;
+
+    if ((ret = clone_tx(tx, &psbt->tx)) != WALLY_OK) {
```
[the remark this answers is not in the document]
achow101: Added

src/psbt.c
achow101: Fixed

src/psbt.c
achow101: Done

include/wally_psbt.h
achow101: Done

src/psbt.c
```diff
@@ -0,0 +1,2120 @@
...
+    }
+    if (final_script_sig) {
+        if (input->final_script_sig) {
+            wally_free(input->final_script_sig);
+            input->final_script_sig = NULL;
+        }
+        if (!clone_bytes(&input->final_script_sig, final_script_sig, final_script_sig_len)) {
+            input->final_script_sig_len = final_script_sig_len;
+        } else {
+            return WALLY_ENOMEM;
```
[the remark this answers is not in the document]
achow101: I'm not sure how to handle that. I wouldn't say it's necessarily bad that the psbt was partially updated.

src/psbt.c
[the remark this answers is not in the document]
achow101: Fixed

src/psbt.c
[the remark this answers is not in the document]
achow101: I've changed this so that it only works on a newly allocated psbt. So there will be no inputs or outputs. If one of these allocations fails, the other one doesn't need to be undone as it is just an allocation of an empty array.

src/psbt.c
achow101: And?

src/psbt.c
achow101: I don't think so. All of the memory allocated is part of the psbt which is passed in by the caller. If these from_bytes functions fail, then the actual `wally_psbt_from_bytes` function will free the entire psbt and thus any memory that was allocated here.

src/psbt.c
achow101: Fixed

src/psbt.c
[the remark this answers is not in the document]
glslang: the usual way this is handled is by allocating what's needed before copying (or assigning) to the final struct. 
if this is going to be the behaviour then it should be documented that these functions may leave the structure in a state that may not be consistent from the point of the view of the caller. don't think we do this elsewhere in wally. in particular we don't free before we know the following allocation will succeed.

src/psbt.c
glslang: if the analysis is correct no need for the if statement

src/psbt.c
glslang: ok

src/psbt.c
jkauffman1: I think ideally any common code would move to shared files. Whether that's internal.c/h or a transaction specific file I am fairly relaxed.

src/psbt.c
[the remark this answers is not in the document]
achow101: I've changed this to allocate to temp pointers before assigning them to the psbt.

src/psbt.c
achow101: Done

src/psbt.c
```diff
@@ -1002,3 +1002,521 @@ static int count_psbt_parts(
...
+    size_t bytes_len,
+    struct wally_psbt **output)
+{
+    const unsigned char *p = bytes, *end = bytes + bytes_len, *key, *value;
+    uint64_t key_len, value_len;
+    uint8_t type;
+    size_t i;
+    int ret;
+    struct psbt_counts *counts = NULL;
+    struct wally_psbt *result = NULL;
```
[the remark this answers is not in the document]
achow101: What would be the benefit of doing that?

src/psbt.c
```diff
@@ -126,3 +126,114 @@ static int add_keypath_item(struct keypath_map *keypaths, struct keypath_item *i
...
+{
+    size_t latest;
+
+    if (sigs->num_items == sigs->items_allocation_len) {
+        size_t new_alloc_len = 1;
+        size_t orig_num_items = sigs->num_items;
+        if (sigs->items_allocation_len != 0) {
+            new_alloc_len = sigs->items_allocation_len * 2;
+        }
+        struct partial_sigs_item *new_items = wally_malloc(new_alloc_len * sizeof(struct partial_sigs_item));
```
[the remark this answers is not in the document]
achow101: In some places, there isn't a `*some_foo` available.

include/wally_psbt.h
```diff
@@ -0,0 +1,134 @@
...
+
+#include "wally_transaction.h"
+#include "wally_core.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+const uint8_t WALLY_PSBT_MAGIC[5] = {'p', 's', 'b', 't', 0xff
+};
```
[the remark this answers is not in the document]
achow101: Fixed

src/psbt.c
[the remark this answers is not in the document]
achow101: I've changed most of these to use `*some_foo`

src/psbt.c
```diff
@@ -237,3 +237,125 @@ static int add_partial_sig_item(struct partial_sigs_map *sigs, struct partial_si
...
+        for (i = 0; i < unknowns->num_items; ++i) {
+            if (unknowns->items[i].key) {
+                wally_clear(unknowns->items[i].key, unknowns->items[i].key_len);
+            }
+            if (unknowns->items[i].value) {
+                wally_clear(unknowns->items[i].value, unknowns->items[i].value_len);
+            }
+        }
+        clear_and_free(unknowns->items, unknowns->num_items * sizeof(*unknowns->items));
+        wally_clear(unknowns, sizeof(*unknowns));
```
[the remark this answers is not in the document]
achow101: Done

src/psbt.c
```diff
@@ -806,3 +806,199 @@ int wally_psbt_set_global_tx(
...
+            }
+            if ((result->input_counts = wally_malloc(result->num_inputs * sizeof(struct psbt_input_counts))) == NULL ||
+                (result->output_counts = wally_malloc(result->num_outputs * sizeof(struct psbt_output_counts))) == NULL) {
+                ret = WALLY_ENOMEM;
+                goto fail;
+            }
+            break;
+        }
+        // Unknowns
+        default: {
```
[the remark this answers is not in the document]
achow101: Fixed

src/psbt.c
```diff
@@ -1002,3 +1002,521 @@ static int count_psbt_parts(
...
+            p += varint_from_bytes(p, &value_len);
+            p += uint64_from_le_bytes(p, &amount);
+            script_len_len = varint_from_bytes(p, &script_len);
+            p += script_len_len;
+            ret = wally_tx_output_init_alloc(amount, p, script_len, &result->witness_utxo);
+            if (ret != WALLY_OK) {
+                return ret;
+            }
+            p += script_len;
+            if (8 + script_len_len + script_len != value_len) {
```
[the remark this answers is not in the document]
achow101: Added

src/psbt.c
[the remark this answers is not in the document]
jb55: otherwise the compiler can't catch when variables are accidently unassigned during refactorings. just a pattern I noticed when working on c-lightning.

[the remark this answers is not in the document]
achow101: Not initializing these seems to be causing a segfault. I think it's because one of them isn't being initialized due to another change I am making, and since the pointer isn't null, something is being freed when it shouldn't be.

src/psbt.c
```diff
@@ -1002,3 +1002,521 @@ static int count_psbt_parts(
...
+            // Read the path length
+            p += varint_from_bytes(p, &value_len);
+            if (value_len % 4 != 0 || value_len == 0) {
+                return WALLY_EINVAL;     // Invalid length for keypaths
+            }
+            path_len = (value_len / 4) - 1;
+
+            // Read the fingerprint
+            memcpy(keypaths->items[keypaths->num_items].origin.fingerprint, p, 4);
+            p += 4;
```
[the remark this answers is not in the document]
achow101: I've added a check after each key and value read that checks that we aren't going to read past the end of the buffer. Instead of an assert, this will return `WALLY_EINVAL`.

stepansnigirev:
A small thing I noticed... Some functions have `wally_` prefix and others don't.
For example `wally_psbt_input_init_alloc(...)` but `add_new_unknown(...)`

src/psbt.c
```diff
@@ -0,0 +1,2188 @@
...
+    struct unknowns_map *unknowns)
+{
+    unsigned char *result_redeem_script = NULL;
+    unsigned char *result_witness_script = NULL;
+    struct keypath_map *result_keypaths = NULL;
+    struct unknowns_map *result_unknowns = NULL;
+    int ret = WALLY_OK;
+
+    if (redeem_script && !clone_bytes(&result_redeem_script, redeem_script, redeem_script_len)) {
+        goto fail;
```
jkauffman1: need to set ret?

src/psbt.c
```diff
@@ -0,0 +1,2188 @@
...
+    }
+    if (witness_script && !clone_bytes(&result_witness_script, witness_script, witness_script_len)) {
+        goto fail;
```
jkauffman1: need to set ret?

src/psbt.c
```diff
@@ -0,0 +1,2188 @@
...
+    struct keypath_map *keypaths,
+    struct unknowns_map *unknowns)
+{
+    unsigned char *result_redeem_script = NULL;
+    unsigned char *result_witness_script = NULL;
+    struct keypath_map *result_keypaths = NULL;
+    struct unknowns_map *result_unknowns = NULL;
+    int ret = WALLY_OK;
+
+    if (redeem_script && !clone_bytes(&result_redeem_script, redeem_script, redeem_script_len)) {
```
jkauffman1: General comment on this function - it may be the case that output already has buffers allocated that are big enough to hold the new redeem_script, witness_script, keypaths and unknowns, in which case it would be better to reuse them rather than freeing the memory and allocating new.

achow101: With how the rest of the functions are written, it's necessary for the buffers to be the exact length of the scripts otherwise too much stuff may be written out. So even if the buffer is big enough, it may not be a good idea to use it.

src/psbt.c
achow101: Done

src/psbt.c
achow101: Done

include/wally_psbt.h
```diff
@@ -0,0 +1,429 @@
...
+#define LIBWALLY_CORE_PSBT_H
+
+#include "wally_transaction.h"
+#include "wally_core.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+const uint8_t WALLY_PSBT_MAGIC[5] = {'p', 's', 'b', 't', 0xff};
```
stepansnigirev: shouldn't it be `extern const` with value itself defined in `psbt.c`?
otherwise if I include `wally_psbt.h` in two different files in the project I got linking problems with `duplicate symbol`...

glslang: your analysis is correct. I suspect we may not want it at all in the header file at this stage either. regardless I'd suggest a function would be preferred as to simplify the bindings for the other languages.

include/wally_psbt.h
```diff
@@ -0,0 +1,429 @@
...
+    unsigned char *redeem_script,
+    size_t redeem_script_len,
+    unsigned char *witness_script,
+    size_t witness_script_len,
+    struct keypath_map *keypaths,
+    struct unknowns_map *unknowns);
+
+#ifndef SWIG_PYTHON
+/**
+ * Free a psbt input allocated by `wally_psbt_output_init_alloc`.
```
glslang: psbt input -> psbt output

include/wally_psbt.h
```diff
@@ -0,0 +1,429 @@
...
+ *
+ * :param input: The psbt input to free.
```
glslang: input -> output
psbt input -> psbt output

include/wally_psbt.h
```diff
@@ -0,0 +1,429 @@
...
+/**
+ * Set the global transaction for a psbt.
+ * Also initializes all of the wally_psbt_input and wally_psbt_outputs necessary
+ *
+ * :param tx: The transaction to set.
+ * :param psbt: The psbt to set the transaction for
+ */
+WALLY_CORE_API int wally_psbt_set_global_tx(
+    struct wally_tx *tx,
+    struct wally_psbt *psbt);
```
glslang: for consistency it'd appear the arguments should be psbt, tx. appears reversed as is.

include/wally_transaction.h
```diff
@@ -3,6 +3,7 @@
 
 #include "wally_core.h"
 #include "wally_crypto.h"
+#include <stdbool.h>
```
glslang: is this include needed here? can't it be added in the src that requires its use?

include/wally_psbt.h
```diff
@@ -0,0 +1,429 @@
...
+};
+#endif /* SWIG */
+
+/**
+ * Allocate and initialize a new keypath map.
+ *
+ * :param alloc_len: The number of items to allocate.
+ * :param output: Destination for the new keypath map
+ */
+WALLY_CORE_API int keypath_map_init_alloc(size_t alloc_len, struct keypath_map **output);
```
glslang: if is part of the public interface then we may want to add wally_ as a prefix to these functions.

include/wally_psbt.h
achow101: I've made it a `extern const`

include/wally_psbt.h
achow101: Fixed

include/wally_psbt.h
achow101: Fixed

include/wally_psbt.h
achow101: Done

include/wally_transaction.h
achow101: Removed. That was from an earlier change.

include/wally_psbt.h
achow101: Changed these functions and the structs to have the `wally_` prefix as they are public.

glslang:
- a simple roundtrip from_base64 to to_base64 leaks memory (there appears to be a few calls where not everything is deallocated)
- PSBT_MAGIC needs to be removed from the header because it interacts badly with visibility hidden meaning it cannot be used at all.
- use memcmp rather than for loops to compare stuff (PSBT magic is one of them. unsure if others exist);
- the update calls appear to have a pattern of if this then that which doesn't appear to be needed and it leads to difficult to read code. I'd suggest
  implementing as how issuance_set is done in transaction.c for example.
- clone_tx needs to use the clone functions as opposed to a roundtrip

achow101:
> a simple roundtrip from_base64 to to_base64 leaks memory (there appears to be a few calls where not everything is deallocated)

Could you help me figure out where those are? I've looked through it myself and did not see anywhere that I thought memory could leak, and valgrind doesn't seem to work, at least not with the python tests.

> use memcmp rather than for loops to compare stuff (PSBT magic is one of them. unsure if others exist);

This was done in the latest push.

glslang:
I used this with valgrind.

```
#include "config.h"

#include <wally_psbt.h>
#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <stdbool.h>

static const char *psbt = "cHNidP8BAHUCAAAAASaBcTce3/KF6Tet7qSze3gADAVmy7OtZGQXE8pCFxv2AAAAAAD+////AtPf9QUAAAAAGXapFNDFmQPFusKGh2DpD9UhpGZap2UgiKwA4fUFAAAAABepFDVF5uM7gyxHBQ8k0+65PJwDlIvHh7MuEwAAAQD9pQEBAAAAAAECiaPHHqtNIOA3G7ukzGmPopXJRjr6Ljl/hTPMti+VZ+UBAAAAFxYAFL4Y0VKpsBIDna89p95PUzSe7LmF/////4b4qkOnHf8USIk6UwpyN+9rRgi7st0tAXHmOuxqSJC0AQAAABcWABT+Pp7xp0XpdNkCxDVZQ6vLNL1TU/////8CAMLrCwAAAAAZdqkUhc/xCX/Z4Ai7NK9wnGIZeziXikiIrHL++E4sAAAAF6kUM5cluiHv1irHU6m80GfWx6ajnQWHAkcwRAIgJxK+IuAnDzlPVoMR3HyppolwuAJf3TskAinwf4pfOiQCIAGLONfc0xTnNMkna9b7QPZzMlvEuqFEyADS8vAtsnZcASED0uFWdJQbrUqZY3LLh+GFbTZSYG2YVi/jnF6efkE/IQUCSDBFAiEA0SuFLYXc2WHS9fSrZgZU327tzHlMDDPOXMMJ/7X85Y0CIGczio4OFyXBl/saiK9Z9R5E5CVbIBZ8hoQDHAXR8lkqASECI7cr7vCWXRC+B3jv7NYfysb3mk6haTkzgHNEZPhPKrMAAAAAAAAA";

#define check_ret(r) if (r != WALLY_OK) return false

static bool tx_roundtrip(const char *tx_hex)
{
    struct wally_psbt *psbt_tx;
    char *new_b64;
    int ret;

    ret = wally_psbt_from_base64(tx_hex, &psbt_tx);
    check_ret(ret);

    ret = wally_psbt_to_base64(psbt_tx, &new_b64);
    if (ret != WALLY_OK || strcmp(tx_hex, new_b64))
        return false;

    ret = wally_free_string(new_b64);
    check_ret(ret);

    wally_psbt_free(psbt_tx);

    return true;
}

static bool test_tx_parse(void)
{
    return tx_roundtrip(psbt);
}

int main(void)
{
    bool tests_ok = true;

#define RUN(t) if (!t()) { printf(#t " test_tx() test failed!\n"); tests_ok = false; }

    RUN(test_tx_parse);

    return tests_ok ? 0 : 1;
}
```

achow101:
>     * a simple roundtrip from_base64 to to_base64 leaks memory (there appears to be a few calls where not everything is deallocated)
Should be fixed now. `wally_psbt_free` and `wally_psbt_input_free` were missing a few things.
>     * PSBT_MAGIC needs to be removed from the header because it interacts badly with visibility hidden meaning it cannot be used at all.
Done
>     * use memcmp rather than for loops to compare stuff (PSBT magic is one of them. unsure if others exist);
Done
>     * the update calls appear to have a pattern of if this then that which doesn't appear to be needed and it leads to difficult to read code. I'd suggest  implementing as how issuance_set is done in transaction.c for example.
So the idea is to instead set everything instead of changing only some parts of the input or output? If so, I don't really think that really fits the design of PSBT. PSBTs are supposed to be updated and modified, not wholesale changed by the caller by a setter. The caller would have to track everything that is already in the input or output and be sure to re-set those things when updating instead of just only specifying the thing to add/change.
>     * clone_tx needs to use the clone functions as opposed to a roundtrip
Done

glslang:
> 
> So the idea is to instead set everything instead of changing only some parts of the input or output? If so, I don't really think that really fits the design of PSBT. PSBTs are supposed to be updated and modified, not wholesale changed by the caller by a setter. The caller would have to track everything that is already in the input or output and be sure to re-set those things when updating instead of just only specifying the thing to add/change.
> 

In which case I need to pass NULL to these functions to mean don't change. This is not the convention in wally. It appears to me all is needed is individual setters and getters then in order to update.

achow101:
> In which case I need to pass NULL to these functions to mean don't change. This is not the convention in wally. It appears to me all is needed is individual setters and getters then in order to update.

I've changed it to use individual setters.

src/psbt.c
```diff
@@ -0,0 +1,2218 @@
...
+    struct wally_psbt_output *output,
+    unsigned char *redeem_script,
+    size_t redeem_script_len)
+{
+    unsigned char *result_redeem_script;
+
+    if (!clone_bytes(&result_redeem_script, redeem_script, redeem_script_len)) {
+        return WALLY_ENOMEM;
+    }
+    if (output->redeem_script) {
```
glslang: this if shouldn't be needed if the initialisation is OK. same for the others. nit: could possibly use the same setup as the TX_SET_B in transaction.c

include/wally_psbt.h
```diff
@@ -0,0 +1,526 @@
...
+ *
+ * :param keypaths: The keypath map to add to
+ * :param pubkey: The pubkey to add
+ * :param fingerprint: The master key fingerprint for the pubkey
+ * :param path: The BIP32 derivation path for the pubkey
+ * :param path_len: The number of items in path
+ */
+WALLY_CORE_API int wally_add_new_keypath(struct wally_keypath_map *keypaths,
+                                   unsigned char pubkey[EC_PUBLIC_KEY_UNCOMPRESSED_LEN],
+                                   unsigned char fingerprint[4],
```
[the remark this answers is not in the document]
achow101: I don't really follow why these need to be a (unsigned char*, size_t) pair. They are fixed sizes, not variable, and the use of the fixed sized pubkey is used elsewhere in the library too.

src/psbt.c
achow101: How so? `output->redeem_script` could always be empty, and AFAICT `wally_free` doesn't check for NULL before freeing.

src/psbt.c
```diff
@@ -0,0 +1,2218 @@
...
+    return WALLY_OK;
+}
+
+int wally_keypath_map_free(struct wally_keypath_map *keypaths)
+{
+    size_t i;
+
+    if (keypaths) {
+        for (i = 0; i < keypaths->num_items; ++i) {
+            if (keypaths->items[i].origin.path) {
```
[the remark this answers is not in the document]
achow101: Done

include/wally_psbt.h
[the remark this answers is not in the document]
glslang: void f(unsigned char a[1000]) == void f(unsigned char *a) == void f(unsigned char[10])

src/psbt.c
glslang: wally_free should forward to free which is a no op for null ptr

include/wally_psbt.h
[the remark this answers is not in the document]
achow101: Done. Not sure about the static stuff, but static is also used elsewhere too.

include/wally_psbt.h
```diff
@@ -0,0 +1,526 @@
...
+/**
+ * Add an item to a partial sigs map
+ *
+ * :param sigs: The partial sigs map to add to
+ * :param pubkey: The pubkey to add
+ * :param sig: The signature to add
+ * :param sig_len: The length of sig
+ */
+WALLY_CORE_API int wally_add_new_partial_sig(struct wally_partial_sigs_map *sigs,
+                                       unsigned char pubkey[65],
```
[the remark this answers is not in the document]
achow101: Done

src/psbt.c
achow101: Done

glslang:
utAck dfa6738676a35e0f8802e08bd12dd5dc3f6cc8d0

jkauffman1:
utAck dfa6738

greenaddress:
merged in 3530c5bfa1d8dfcd3b5f5e4886f14036de303c8e

PR #127, opened 2019-08-25
docs: Add list of projects and products that are known to depend on or use libwally

practicalswift:
Add list of projects and products that are known to depend on or use `libwally`.

Am I missing any known users? :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,3 +154,22 @@ $ ./tools/coverage.sh
 
 The coverage report can then be viewed at `src/lcov/index.html`. Patches to
 increase the test coverage are welcome.
+
+## Users of libwally-core
+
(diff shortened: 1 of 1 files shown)


glslang:
utAck 471902c

greenaddress:
utack 471902c, my only nit is that I suggest we put a little disclaimer that those projects may be experimental, superseded etc

practicalswift:
@greenaddress Good point! I've now added a disclaimer -- please re-review.

Also: am I missing any prominent users of the library?

greenaddress:
@practicalswift looks good to me, thank you!

I checked with a few and we think you got all prominent users of the library, cheers, will merge

PR #133, opened 2019-09-10
configure: update pkgconfig version for next release

jb55:
I noticed the pkgconfig version was reporting 0.7.2 instead of 0.7.4
when packaging wallycore in nix. It is most likely because of this
version which was not updated.

Ideally we would have this auto-update somehow, but for now let's just
prepare it for the next release.

Signed-off-by: William Casarin <jb55@jb55.com>


greenaddress:
merged in f80018570826bb542c388986499f205a06f38e64 in master

thanks! closing

PR #147, opened 2019-10-10
bip32: add key_strip_private_key to the public api

benma:
It is useful for when you want an xpub at a keypath like m/44'/0'/0'.

### DIFF
--- a/include/wally_bip32.h
+++ b/include/wally_bip32.h
@@ -298,6 +298,16 @@ WALLY_CORE_API int bip32_key_from_base58_alloc(
     const char *base58,
     struct ext_key **output);
 
+/**
+ * Converts a private extended key to a public extended key. Afterwards, only public child extended
+ * keys can be derived, and only the public serialization can be created.
+ * If the provided key is already public, nothing will be done.
+ *
+ * :param hdkey: The extended key to covert.
+ */
+WALLY_CORE_API int bip32_key_strip_private_key(
+    struct ext_key *hdkey);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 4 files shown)


benma:
If this is okay, I would also like to add functions `bip32_key_is_public()` and `bip32_key_is_private()`. The only way to figure out what type it is currently is to try to serialize it with the wrong flag, or inspecting `key.priv_key[0]`, which I assume should be treated as an internal detail.

include/wally_bip32.h
```diff
@@ -298,6 +298,16 @@ WALLY_CORE_API int bip32_key_from_base58_alloc(
...
     struct ext_key **output);
 
+/**
+ * Converts a private extended key to a public extended key. Afterwards, only public child extended
+ * keys can be derived, and only the public serialization can be created.
+ * If the provided key is already public, nothing will be done.
+ *
+ * :param hdkey: The extended key to covert.
+ */
+WALLY_CORE_API void bip32_key_strip_private_key(
```
glslang: should return int as per all of the api calls.

src/bip32.c
```diff
@@ -735,6 +735,11 @@ int bip32_key_from_base58_alloc(const char *base58,
     return ret;
 }
 
+void bip32_key_strip_private_key(struct ext_key *hdkey)
+{
+    key_strip_private_key(hdkey);
```
glslang: it should check for a null pointer and bail out.

include/wally_bip32.h
benma: I see, that was also the reason for the swig fail. Thanks.

src/bip32.c
benma: Done

greenaddress:
utack dbfd4d09b4b89f5cee25a8bbbcfb4aac74db242f merged (rebased) in master ias ff01f0749006161e25363234bcf9bad7da60f985 - thanks!

PR #150, opened 2019-11-18
msvc build support.

k-matsuzawa:
1. Repair Windows build errors due to ssize_t.
- base64.h: add config.h, stdint.h
- windows_config/config.h: add ssize_t define

2. Update msvc/build.bat, gen_ecmult_static_context.bat, and wallyjs.bat
- add target newer source code
- add UTF-8 option
- add define: ECMULT_WINDOW_SIZE
- add elements build selecting

### DIFF
--- a/src/ccan/ccan/base64/base64.h
+++ b/src/ccan/ccan/base64/base64.h
@@ -2,6 +2,8 @@
 #ifndef CCAN_BASE64_H
 #define CCAN_BASE64_H
 
+#include "config.h"
+#include <stdint.h>
 #include <stddef.h>
 #include <stdbool.h>
 #include <sys/types.h>
--- a/src/wrap_js/windows_config/binding.gyp.elements_tmpl
+++ b/src/wrap_js/windows_config/binding.gyp.elements_tmpl
@@ -1,0 +1,37 @@
+{
+  "targets": [
+    {
+      "target_name": "deps",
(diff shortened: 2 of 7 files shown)


src/ccan/ccan/base64/base64.h
```diff
@@ -2,6 +2,8 @@
 #ifndef CCAN_BASE64_H
 #define CCAN_BASE64_H
 
+#include "config.h"
+#include <stdint.h>
```
glslang: nit: in the unlikely event we need to update this file then theses changes may get lost. unsure if there's a way of enforcing this in the build somehow.

glslang:
utAck  abb85a2

src/ccan/ccan/base64/base64.h
k-matsuzawa: > in the unlikely event we need to update this file then theses changes may get lost.

due to the current structure of this project, it is necessary to include `config.h` (or libsecp256k1-config.h) at the top of the file.
This is done for the other `ccan` header files as well.
As an easy way, can create a tool that checks `#include "config.h"` at building. (Error if not exist)
Auto-generation of source code's can solve the problem, but it takes too much time.
As long as you use only `autotools`, MSVC build is difficult ...

greenaddress:
merged, in master as of db00332cf2fa426285480a2f8d45ef867c0838a5
(had to rebase)

PR #156, opened 2019-12-03
Add flag to sort multisig keys (BIP67)

Sjors:


Sjors:
I can't make sense of these Travis failures:
```
swig_java/src/com/blockstream/libwally/Wally.java:645: error: incompatible types: byte[] cannot be converted to int
2297      final int len = scriptpubkey_multisig_from_bytes(bytes, threshold, flags, buf);
```

glslang:
you removed the const from the function signature. I'd prolly do a copy and sort that instead.

Sjors:
Done

src/script.c
```diff
@@ -541,34 +541,46 @@ int wally_scriptpubkey_p2sh_from_bytes(
...
+}
+
 int wally_scriptpubkey_multisig_from_bytes(
     const unsigned char *bytes, size_t bytes_len, uint32_t threshold,
     uint32_t flags, unsigned char *bytes_out, size_t len, size_t *written)
 {
     size_t n_pubkeys = bytes_len / EC_PUBLIC_KEY_LEN;
     size_t script_len = 3 + (n_pubkeys * (EC_PUBLIC_KEY_LEN + 1));
     size_t i;
+    unsigned char pubkey_bytes[bytes_len];
```
glslang: use of VLA is causing build failures and we want to avoid them in the codebase.

src/script.c
```diff
@@ -541,34 +541,46 @@ int wally_scriptpubkey_p2sh_from_bytes(
     return ret;
 }
 
+int pubkey_compare(const void *a, const void *b)
```
glslang: should be declared static

src/script.c
```diff
@@ -541,34 +541,46 @@ int wally_scriptpubkey_p2sh_from_bytes(
...
+    unsigned char pubkey_bytes[bytes_len];
 
     if (written)
         *written = 0;
 
     if (!bytes || !bytes_len || bytes_len % EC_PUBLIC_KEY_LEN ||
         n_pubkeys < 1 || n_pubkeys > 16 || threshold < 1 || threshold > 16 ||
-        threshold > n_pubkeys || flags || !bytes_out || !written)
+        threshold > n_pubkeys || (flags & ~WALLY_SCRIPT_MULTISIG_SORTED) ||
+        !bytes_out || !written)
```
glslang: !written may  be incorrect since we test it above and bypass written if is NULL

src/script.c
Sjors: I changed it to the maximum size needed for 16 pubkeys.

src/script.c
Sjors: I didn't touch that, but they way I see it, the _value_ of written is set to `0` if the _pointer_ exists. The check here cares if the pointer exists. It might make more sense to move the `*written = 0` below this line.

src/script.c
Sjors: done

src/script.c
glslang: you're right.  we can just leave it unless you want to change.

Sjors:
Travis machines 4 and 10 are still failing. Hopefully unrelated, but it would be nice if they produced more useful error messages.

glslang:
utAck 0355c512556009d740dc49c7809b75b147392383

greenaddress:
merged in f4364a9f77edc95ce49e4638809e0bbd01a5ba9a in master - thanks!

PR #158, opened 2019-12-16
(wrap_js) Add a guard to an empty LocalBuffer

k-matsuzawa:
If an empty buffer is set to LocalBuffer, set a dummy empty buffer to `LocalBuffer::mData` .

`LocalBuffer::mData` is set by `node::Buffer::Data()` .
However, `node::Buffer::Data()` is an API that calls nodejs directly, and its behavior has changed in nodejs v13.3.0.
Currently, `node::Buffer::Data()` on nodejs v13.3.0 has get null pointer from empty LocalBuffer.

Therefore, set the pointer address of an empty buffer so that a null pointer is not passed.

### DIFF
--- a/src/wrap_js/makewrappers/templates/nan.py
+++ b/src/wrap_js/makewrappers/templates/nan.py
@@ -32,13 +32,13 @@ static bool IsValid(const typename Nan::Maybe<T>& maybe)
 // Binary data is expected as objects supporting the JS Buffer interface
 struct LocalBuffer {
     LocalBuffer(Nan::NAN_METHOD_ARGS_TYPE info, int n, int& ret)
-        : mData(0), mLength(0)
+        : mData(0), mLength(0), mDummyData(0)
     {
         Init(info[n], ret);
     }
 
     LocalBuffer(const v8::Local<v8::Value>& obj, int& ret)
-        : mData(0), mLength(0)
+        : mData(0), mLength(0), mDummyData(0)
     {
         Init(obj, ret);
     }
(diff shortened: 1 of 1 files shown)


glslang:
It'd appear to me if  is returning a null pointer then the code should be calling IsValid and set ret to WALLY_ERROR or WALLY_EINVAL.

k-matsuzawa:
Thanks!
All that was needed was to replace NULL with an empty buffer in the normal route, so it was moved the position.

glslang:
Don't understand what you mean. My comment still stands AFAICS.

k-matsuzawa:
Supplement the explanation.

Cause:
- After updated to node.js v13, wrap_js test fail on test_aes.js and test_hash.js.
- An error occured API is wally_sha256, wally_aes_cbc.
- node.js's API (node::Buffer::Data()) has been changed.
  under node.js v12, `node::Buffer::Data()` is return not NULL on empty buffer.
  but, node.js v13 latest, `node::Buffer::Data()` is return NULL on empty buffer.
- A NULL pointer was specified for the API, resulting in an error.

Reason:
- It is as specified by the C language API that an error occurs when buffer is set NULL pointer.
  The problem lies in the wrap_js code.
- By not specifying null on node.js, I decided to avoid error.
- It does not affect other languages. (C, Python, Java)

glslang:
It appears to me all wally functions that can take a length of zero like aes_cbc need to account for the fact that bytes can also be NULL in that case and be made safe (as in bytes can never be dereferenced). This would be the case for the encrypt case in aes_cbc.

k-matsuzawa:
1f716419 Thanks!
Cancel this request.

PR #162, opened 2019-12-21
Add wally_scriptpubkey_to_address

### LINKED ISSUE #118: Infer address from scriptPubKey
The user would start with [wally_scriptpubkey_get_type](https://wally.readthedocs.io/en/release_0.7.0/script/#c.wally_scriptpubkey_get_type). This returns the `WALLY_SCRIPT_TYPE_` type. Then we could add individual parser functions for each type, or one generic one:

```c
scriptpubkey_to_address(const unsigned char *bytes, size_t bytes_len, \
                        uint32_t scriptpubkey_type, unsigned char *bytes_out, size_t len)
```

When `scriptpubkey_type` is 0 it infers the type.

For SegWit this is uncessary; `wally_addr_segwit_from_bytes` works fine.

Sjors:
Closes #118 

The new `wally_scriptpubkey_to_address` function could be expanded later to also work with SegWit, but I can't be bothered :-)

### DIFF
--- a/include/wally_address.h
+++ b/include/wally_address.h
@@ -88,6 +88,22 @@ WALLY_CORE_API int wally_address_to_scriptpubkey(
     size_t *written);
 
 /**
+ * Infer address from a scriptPubKey. For SegWit addresses, use `wally_addr_segwit_from_bytes`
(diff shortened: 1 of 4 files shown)


glslang:
utAck 2955e956f62d253e0f03f1455554cd0fb5ffbf56

PR #169, opened 2020-01-23
script: fail scriptpubkey_multisig_from_bytes on too short out buf

benma:
I accidentally called wally_scriptpubkey_multisig_from_bytes with the
outbuffer size set to 0. To my surprise, the function returned
WALLY_OK, with the `written` var set to the expected script size. In
other words it looked like the full script has been written to the
output buffer successfully even though nothing has happened.

This should obviously fail instead.


benma:
I was informed that I might've missed the case that the function can be used to figure out the script len by providing `0` as the output buffer length. Is that the intended use of this?

If so I could change the PR to document this case and that a `if (written > outlen) { failure }` check is needed after calling the function.

glslang:
To my knowledge the original author intended that to be the case. This is the case in many other places although some other functions diverged from it.

benma:
Thanks. Opening a new PR shortly to improve the docs.

PR #173, opened 2020-01-31
Use absolute path to dynamic library on macos

philippem:
- make check was failing on macOS Catalina with the following error:

  File "/Users/phil/src/github.com/philippem/libwally-core/src/test/util.py", line 15, in <module>
    libwally = CDLL(root_dir + 'src/.libs/libwallycore.' + SO_EXT)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ctypes/__init__.py", line 366, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(../src/.libs/libwallycore.dylib, 6): no suitable image found.  Did find:
	file system relative paths not allowed in hardened programs

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Please report bugs and submit patches to https://github.com/ElementsProject/libw
 Wally can currently be built for:
 - Linux
 - Android
-- OS X
+- macOS
 - iOS
 - Windows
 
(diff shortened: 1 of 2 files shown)


greenaddress:
@philippem on what version of osx is this needed? It seems on 10.15.2 isn't needed, is this for 10.15.3 or?

philippem:
It is not required for 10.13.6. It is required for 10.15.3. Not sure about versions in between.
I've updated the README in a subsequent amendment to the commit.

greenaddress:
utack 26bc93f5c522177c17662b73058bc929962fb601

greenaddress:
merged in master in 4040573055f5151777b0315795ebc1e98b7a74f6

Issue #176, opened 2020-02-17
Fails to compile with gcc 4.8


rustyrussell:
I'm surprised it works with any gcc, in fact:

```
  CC                      = gcc-4.8
  CFLAGS                  = -g -W -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wno-unused-function -Wno-long-long -Wno-overlength-strings -fvisibility=hidden -O3
```

Note the `-std=c89` (which AFAICT comes from libsecp):

Sure enough, libwally is not C89.  For example:

```
../../libwally-core/src/address.c: In function ‘wally_bip32_key_to_address’:
../../libwally-core/src/address.c:25:5: warning: C++ style comments are not allowed in ISO C90 [enabled by default]
     // Catch known incorrect combinations of address type and version:
     ^
../../libwally-core/src/address.c:25:5: warning: (this will be reported only once per input file) [enabled by default]
../../libwally-core/src/elements.c: In function ‘wally_asset_pak_whitelistproof’:
../../libwally-core/src/elements.c:629:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (size_t i = 0; i < num_keys; ++i) {
     ^
../../libwally-core/src/elements.c:629:5: note: use option -std=c99 or -std=gnu99 to compile your code
../../libwally-core/src/sign.c: In function ‘wally_ec_sig_to_public_key’:
../../libwally-core/src/sign.c:324:5: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
     int recid = (sig[0] - 27) & 3;
     ^
```

glslang:
Should compile now with -std=c89. Please do let us know if you're still seeing errors.

vasild:
FWIW the above mentioned CFLAGS that include an explicit `-std=c89` are only used to compile `src/secp256k1` which is C89 compliant and this is fine.

libwally's own source files are compiled without an explicit `-std=`. For an exact compilation command see the commit message of https://github.com/ElementsProject/lightning/pull/3561/commits/6395c4340b6efe5a463fbf2d47b76726e62a7041.

Recent compilers default to C99 or newer, but gcc 4.8 [defaults to gnu89](https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/C-Dialect-Options.html#C-Dialect-Options).

This also fixes the issue:
```diff
--- i/configure.ac
+++ w/configure.ac
[...]

PR #177, opened 2020-02-20
update for c89 compatibility (compiled against gcc-4.8 on debian jessie)

glslang:
should address #176


greenaddress:
utack 06522eb55655bceee39a27b8a91f0e9fef73aec1

PR #186, opened 2020-05-15
Psbt extensions and elements psbt serialization

achow101:
Implements the latest additions to BIP 174 (version field, types are compact size unsigned int, proprietary types) as well as the Elements PSBT proprietary fields.


achow101:
Someone who is more familiar with elements should probably add the elements specific things to `wally_sign_psbt` and `wally_finalize_psbt` and whatever other functions are needed to use elements psbt. This PR is primarily serialization.

Additionally, it is not clear to me how to test these changes as elements doesn't have serialization tests like there are for BIP 174.

achow101:
Redoing this on top of #191

PR #190, opened 2020-05-19
clone_bytes: set *dst to NULL if we are asked to clone a zero length value

rustyrussell:
This fixes psbt: in many places it assumes this behavior.  Older code carefully initializes to NULL before calling clone_bytes, so it's not broken by this change, but it's deeply surprising to new authors!

In particular, valgrind noticed the unitialized var in line 1665 in psbt.c:

	if (input->final_script_sig) {

This is set in wally_psbt_input_set_final_script_sig:

    unsigned char *result_final_script_sig;

    if (!clone_bytes(&result_final_script_sig, final_script_sig, final_script_sig_len)) {
    ...
    input->final_script_sig = result_final_script_sig;

Surprise, input->final_script_sig is now random garbage from the stack.

PSBT assumes similar behavior in other places: it seems easiest to fix the clone_bytes footgun rather than change all the psbt code.

Reported-by: @niftynei
Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

### DIFF
--- a/src/transaction.c
+++ b/src/transaction.c
@@ -157,8 +157,10 @@ static bool is_valid_elements_tx(const struct wally_tx *tx)
 
 bool clone_bytes(unsigned char **dst, const unsigned char *src, size_t len)
 {
-    if (!len)
+    if (!len) {
+	*dst = NULL;
         return true;
+    }
     *dst = wally_malloc(len);
     if (*dst)
         memcpy(*dst, src, len);
--- a/src/transaction.c
+++ b/src/transaction.c
@@ -158,7 +158,7 @@ static bool is_valid_elements_tx(const struct wally_tx *tx)
 bool clone_bytes(unsigned char **dst, const unsigned char *src, size_t len)
 {
     if (!len) {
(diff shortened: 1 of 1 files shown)


jkauffman1:
utack 13604d49c792973d507bca737403684316a64e7b

src/transaction.c
```diff
@@ -157,8 +157,10 @@ static bool is_valid_elements_tx(const struct wally_tx *tx)
 
 bool clone_bytes(unsigned char **dst, const unsigned char *src, size_t len)
 {
-    if (!len)
+    if (!len) {
+	*dst = NULL;
         return true;
```
LeoComandini, line 162: nit ([needed for CI](https://github.com/ElementsProject/libwally-core/blob/master/.gitlab-ci.yml#L14-L15)):
```suggestion
        *dst = NULL;
        return true;
```

LeoComandini:
utack bb2e7d1779ba0e54c7a0037e7d3ddedccddb90b5

PR #191, opened 2020-05-20
Fix PSBT marshal/unmarshal functions.

rustyrussell:
Started with PSBT, will continue.  But worth reviewing.


src/pullpush.c
```diff
@@ -0,0 +1,153 @@
...
+    if (cursor == NULL || *cursor == NULL) {
+        *max += len;
+        return NULL;
+    }
+    if (len > *max) {
+        if (src)
+            memcpy(*cursor, src, *max);
+        /* From now on, max records the room we *needed* */
+        *max = len - *max;
+        *cursor = NULL;
```
achow101, line 21: What's the purpose of doing this on failure?

achow101:
Concept ACK

src/pullpush.c
rustyrussell: *cursor == NULL is the way you detect unmarshalling failure.

*max increasing is how you tell how many bytes short you are.

You'll see this explicitly used here:

```
    if (cursor == NULL) {
            /* Once cursor was NULL, max accumulates hm bytes we needed */
            *bytes_written = len + max;
            return WALLY_EINVAL;
    } else {
            *bytes_written = len - max;
    }
```

rustyrussell:
This still fails travis, but I can't tell *why* since the logs are insufficient :(

glslang:
Look at tools/travis_build.sh and add your new test to it. you should have the output of it somewhere in the build output

rustyrussell:
> Look at tools/travis_build.sh and add your new test to it. you should have the output of it somewhere in the build output

OK, I fixed it a different way in the first commit: let's see how that looks?

glslang:
Nice. #197 fixes the actual travis build so hopefully all current outstanding PRs will have a green build.

rustyrussell:
> A few comments
> 
>     * In [8484e97](https://github.com/ElementsProject/libwally-core/commit/8484e9760d89419fbd56e796685569150c7b6b36) "travis: don't explicitly list what tests in tools/travis_build.sh": Maybe don't @ mention people in commit messages. In Core, we have a policy that commit messages shouldn't do that as those people will get spammed when forks pick up those commits.

Ah, I gave up reading my GH notifications long ago, so I never noticed.  Good idea, thanks!

>     * In [637e4c5](https://github.com/ElementsProject/libwally-core/commit/637e4c58a5f355c2d6a238cdf31676dd1dd2149e) "psbt: add a simple test case, by copying some of BIP 0174.": `src/test_psbt*` needs to be added to `.gitignore`.

Ack, and same with limits test.

>     * In [e06ad79](https://github.com/ElementsProject/libwally-core/commit/e06ad79ea226348dcc7ef960fea88d475723ff97) "pullpush: add routines for safe marshal / unmarshal.": There are a few places where `max` and `cursor` are dereferenced without checking that they are not null.. This seems okay to me as the pushpull functions are internal and we would want to see a segfault as that indicates we are using them improperly. But maybe that's not the intention?

Yes, they're internal only, and have specific semantics; adding checks like this would seem to add nothing but clutter.  In particular, there's no infra in libwally for "crash iff in dev mode" which would allow CI to be effective in this case. If we were to return WALLY_EINVAL on internal errors, we might not notice them in CI.

rustyrussell:
Rebased with fixes from @achow101 (all minor).

achow101:
ACK 1cccf0581cba859d5da36eadca7ffed2eb2cf070

achow101:
@glslang @greenaddress Can this be merged?

jgriffiths:

PR #192, opened 2020-05-21
psbt: pass in NULL for script if script_len == 0

niftynei:
`wally_tx_output_init_alloc` will fail if the script_len == 0 but the
buffer passed in for the script is not NULL; consequently if
`script_len` is 0 we should pass in a NULL pointer for the script.

### DIFF
--- a/src/psbt.c
+++ b/src/psbt.c
@@ -1120,7 +1120,7 @@ static int psbt_input_from_bytes(
             p += uint64_from_le_bytes(p, &amount);
             script_len_len = varint_from_bytes(p, &script_len);
             p += script_len_len;
-            ret = wally_tx_output_init_alloc(amount, p, script_len, &result->witness_utxo);
+            ret = wally_tx_output_init_alloc(amount, script_len ? p : NULL, script_len, &result->witness_utxo);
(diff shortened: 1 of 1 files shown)


niftynei:
on second thought, maybe a failure here is desirable behaviour? the current c-lightning implementation abuses the witness-utxo field as a placeholder for the input amount, so in some cases the script is set to NULL.

glslang:
see BYTES_VALID in transaction.c. that's the usual check in wally. an error is preferable.

rustyrussell:
> see BYTES_VALID in transaction.c. that's the usual check in wally. an error is preferable.
> EDIT: it's possible psbt is missing thischeck in many other places.

Yes, see #191

niftynei:
> see BYTES_VALID in transaction.c.

Yes I have seen BYTES_VALID, that's the motivation for this patch. :laughing:

rustyrussell:
Hmm, in this case, it's invalid to have a zero-length witness script, so the return is correct (note that #191 has this exact same behavior, despite reworking this code).

But that's actually indep of whether script is NULL or not (as it is in this path).

jgriffiths:

PR #195, opened 2020-05-23
feat: add WALLY_DISABLE_OP_CODE guard

k-matsuzawa:
libwally-core defines OP_CODE using define.
Due to that, the name of OP_CODE cannot be defined in the library being used. (The macro overwrites everything)
The OP_CODE name is a special word in Bitcoin scripts, so we want to use the same name in our Bitcoin application or library. In that case, defining with define is not desirable.
As a countermeasure, we want to be able to suppress the OP_CODE definition of libwally-core in the application or library to be used. It does not affect the build of libwally-core.

### DIFF
--- a/include/wally_script.h
+++ b/include/wally_script.h
@@ -37,6 +37,7 @@ extern "C" {
 #define WALLY_SCRIPT_MULTISIG_SORTED  0x8 /** Sort public keys (BIP67) */
 
 /* Script opcodes */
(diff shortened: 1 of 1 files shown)


LeoComandini:
edf05143a2bb5769f28bc74a5646fa1c15025230 fixed the travis build, you may want to rebase your merge requests to have a green build.

jgriffiths:
Thanks, this will be merged with the review_updates branch PR shortly.

PR #196, opened 2020-05-23
fix: mnemonic memory leak.

k-matsuzawa:
A memory leak was found during a memory leak test in a library that uses libwally-core.
The outline of the modification is as follows.

src/wordlist.c: wordlist_free
  - if check `w->str_len != 0`, no call `wally_free(w)` case.
  - remove top `w->str_len != 0` check, and if `w->str_len != 0` call wally_clear before wally_free.

src/wordlist.c: wordlist_alloc
  - remove `w->str_len = 0;`
  - instead of, insert wally_clear(w, sizeof(*w)). (for struct initializing)

src/mnemonic.c: mnemonic_to_bytes
  - Move the call to `wordlist_init`.
    - because `if (!w || !bytes_out || !len)` case is not call wordlist_free(mnemonic_w).


jgriffiths:
Will be merged with #210 - thanks!

PR #198, opened 2020-05-29
psbt: allow a caller to set the final-witness stack to null

niftynei:
mimics the functionality of `wally_tx_input_set_witness` -- we need this
for c-lightning so that we can unset the witness data for a transaction
before serializing the psbt to disk.


niftynei:
the failure on this looks spurious, probably needs to be re-run https://travis-ci.org/github/ElementsProject/libwally-core/jobs/692405120

jgriffiths:

PR #200, opened 2020-05-29
psbt: fix broken script finder in wally_finalize_psbt

niftynei:
As written, this was looking at the output script on the global_tx for
an input on the same, which caused valgrind problems because A) there's
no guarantee that an output at that input index exists and also B)
that's not even the right script.

What you *want* to be doing is using the non_witness_utxo to look up the
script iff it exists.

If there's no out_script available we totally bail on finalizing this
input at all.


LeoComandini:
utack 5642664b6159b0fe3940a1276f23998dcfce5481

PR #201, opened 2020-05-30
PSBT Extensions and Elements PSBT Serialization

achow101:
Implements the extensions to BIP 174 which add a global version field and proprietary types. The elements proprietary fields are implemented here too.

The elements PSBT fields are added to the psbt structs as well as their serialization and deserialization of those fields. None of the roles for elements psbt have been implemented as what they are supposed to do is still being worked out.

Based on #191


jgriffiths:

PR #203, opened 2020-06-05
Fixup borked PSBT signing + finalize

niftynei:
Found ~two~ three problems with the PSBT sign + finalize code; this PR contains suggested fixes for them and tests.

- Disallow creating a second partial_sig if there's already a sig for that pubkey attached to an input. Creating PSBTs with duplicate key'd partial_sig maps fails to _decode_ from b64.
-  Fixup the out-script logic for finalize; as written fails to properly detect p2sh-p2wpkh outputs
- Populate p2sh-p2wpkh script-sig from the provided redeemscript

Note: disallowing duplicate key'd partial_sigs made the invalid_sigs tests start to pass since the existing test already had partial_sig data populated for the given key; we revert them to 'failing' by removing the existing partial_sig data from the test case.


src/psbt.c
```diff
@@ -2716,28 +2716,42 @@ int wally_finalize_psbt(struct wally_psbt *psbt)
         struct wally_psbt_input *input = &psbt->inputs[i];
         struct wally_tx_input *txin = &psbt->tx->inputs[i];
         unsigned char *out_script = NULL; /* Script that determines how we should finalize this input, typically output script */
-        size_t out_script_len, type;
+        /* Expanded script, only applicable for P2SH variants. Used for determining if is multisig */
+        unsigned char *script = NULL;
```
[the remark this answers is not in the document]
achow101: IMO keeping it `out_script` makes it more readable as `out_script` is supposed to be the scriptcode and scriptcode is what matters, regardless of the 'levels'. It seems like the root cause is setting `out_script` in the wrong order.

src/psbt.c
```diff
@@ -2568,6 +2568,24 @@ int wally_sign_psbt(
             continue;
         }
 
+        /* Make sure we don't already have a sig for this input ?! */
+        if (input->partial_sigs) {
+            for (j = 0; j < input->partial_sigs->num_items; j++) {
+                struct wally_partial_sigs_item *item = &input->partial_sigs->items[j];
+                if ((item->pubkey[0] == 0x04
```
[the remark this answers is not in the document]
achow101: Yes

src/psbt.c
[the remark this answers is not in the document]
niftynei: i'm going to agree to disagree with you here. if reusing `out_script` is necessary to get this PR approved, i will update it to reuse `out-script`.

niftynei:
hey @achow101 i've made the changes you requested, would you mind taking another look? this is currently blocking some other work.

src/psbt.c
```diff
@@ -2737,18 +2737,13 @@ int wally_finalize_psbt(struct wally_psbt *psbt)
             out_script_len = out.script_len;
         }
         if (input->witness_script) {
-            script = input->witness_script;
-            script_len = input->witness_script_len;
+            out_script = input->witness_script;
+            out_script_len = input->witness_script_len;
             witness = true;
         }
         if (input->redeem_script) {
```
achow101, line 2744: This if block can be swapped with the one above. Then the `if (type == WALLY_SCRIPT_TYPE_P2WSH...)` block can be dropped.

src/psbt.c
```diff
@@ -2726,7 +2726,7 @@ int wally_finalize_psbt(struct wally_psbt *psbt)
         }
 
         /* Note that if we patch libwally to supply the non-witness utxo tx field (tx) for
-         * witness inputs also, we'll need a different way to signal p2wpkh scripts */
+        * witness inputs also, we'll need a different way to signal p2sh-p2wpkh scripts */
```
achow101, line 2729: nit: indentation.

niftynei:
squashed + updated

achow101:
ACK f727cbedcac3ab60d3226dd6692ae7501516e37f

src/psbt.c
niftynei: turns out `uncrustify` wants it this way, so reverted this suggested change.

achow101:
ACK e0d0634aea716d813744326ea6c7590eb9fc381c

src/psbt.c
```diff
@@ -2757,6 +2783,16 @@ int wally_finalize_psbt(struct wally_psbt *psbt)
                 if ((ret = wally_witness_p2wpkh_from_der(partial_sig->pubkey, pubkey_len, partial_sig->sig, partial_sig->sig_len, &input->final_witness)) != WALLY_OK) {
                     return ret;
                 }
+                if (input->redeem_script) {
+                    /* P2SH wrapped witness requires final scriptsig of pushing the redeemScript */
+                    script_sig_len = varint_get_length(input->redeem_script_len) + input->redeem_script_len;
+                    input->final_script_sig = wally_malloc(script_sig_len);
```
jkauffman1, line 2789: May be worth explicitly checking the value returned by wally_malloc for null and returning WALLY_ENOMEM

niftynei: this is copied from https://github.com/ElementsProject/libwally-core/blob/master/src/psbt.c#L2860-L2868, which will also need to be cleaned up

niftynei: in the interest of getting my PRs in c-lightning unblocked, we should merge this PR and clean up the malloc problems in a second fixup

niftynei: ok, i went through and updated all `wally_alloc` calls in psbt.c that were missing a NULL check, see #204

PR #207, opened 2020-07-07
tests: enable non swig elements tests even when swig not enabled

niftynei:
too many if nests meant that the build_elements tests would only run if
both elements + swig_python were enabled. here we split this into two
sections, where non-swig build_elements tests will run iff build_elements is
flagged.


jgriffiths:

PR #208, opened 2020-07-10
c-lightning blessed branch

niftynei:
Set of patches used to get PSETs (and, minorly, partial txs) working in c-lightning. Posted here as reference.

Combines patches from #206, #201, and #191, as well as another set of patches to fix found errors in the outstanding PSET serialization PR.

PR #206, opened 2020-07-02
psbt: allow partial transactions.

rustyrussell:
During PSBT preparation, it can reasonably contain unfinished transactions.
In particular, we didn't allow marshalling of zero-output transactions.

Fixes: #205

PR #204, opened 2020-06-22
psbt: return ENOMEM for any wally_malloc failures

niftynei:
A few wally_malloc's were missing a NULL/ENOMEM check

PR #199, opened 2020-05-29
readme: add easier to skim notes for running the tests + formatter

niftynei:
Make it easier to skim the docs to find relevant notes for running the
formatter and test suite

PR #210, opened 2020-07-14
Review updates

### LINKED ISSUE #139: Some `wally_tx_*` calls overreact if the caller tries to be helpful and hints with `WALLY_TX_FLAG_USE_ELEMENTS`
Some calls, e.g., `wally_tx_get_length`, will fail if we specify that we use elements when calling them, leading the caller to have to mask out the bits in the call.

An example of this can be found here:

https://github.com/ElementsProject/lightning/blob/22ea458298238dcf23dfb1dffae9ad8b9725b94b/bitcoin/tx.c#L325-L329

It'd be nice to perform this masking inside the function itself, in order to expose a more consistent interface. It could also verify that the flag matches with the internal detection of elements transactions and return an error if they don't match (though that'd be a breaking API change).

### LINKED ISSUE #184: Missing wally_tx_clone() function.
This is useful, and we want it for c-lightning!

Please expose the internal one!

(You should also create wally_psbt_clone(), too...)

### LINKED ISSUE #189: wally_psbt_set_global_tx, wally_tx *tx should be const
The second parameter, `struct wally_tx *tx` for method `wally_psbt_set_global_tx` should be const, since it's copied internally and not modified.

### LINKED ISSUE #205: psbts with 0 outputs should be serializable.
We are using psbts to hand around our new APIs for merging later.  It's particularly convenient for our new "fundpsbt" to return a set of utxos to spend in psbt form.

Adding a dummy output is awkward: we have have overfunded and the user may want to create a change output, so we'd rather them end up with an invalid tx than accidentally pay a huge fee.

Unfortunately, transaction.c:1986 refuses to allow a zero output tx, so converting our psbt to base64 fails...

jgriffiths:
A large set of fixes and updates encompassing:

    Review and bug fixing of changes over the last year
    Documentation improvements
    Standardising of interfaces and consistency of implementation
    Improved tests
    Merging all outstanding PSBT changes required for current c-lightning master
    Additional PSBT fixes and cleanups
    Wrapper updates

Although there remains a significant amount of PSBT work to complete, I think this is worth merging now so that the lightning peeps can move back to wally master instead of their hacked/blessed branch, and further changes can be smaller and more easily reviewable.

Obsoletes or includes the following PRs: #208 #207 #206 #204 #201 #199 #198 #194 #193 #191

Fixes #205
Fixes #189
Fixes #184
Fixes #139


CHANGES.md
```diff
@@ -1,50 +1,58 @@
...
 ## Version 0.7.8
 
-- Deprecate Python 2 wheels
-- Honor CFLAGS, default to O2
-- Bug fixing
+- Python 2 wheels are now deprecated. Users should move to Python 3 as soon as possible.
 
 ## Version 0.7.7
 
-- Expose wally_psbt_get_length
```
LeoComandini, line 11: Why CHANGES.md should not include API changes that does not break the API?
I would expect to find the addition of new exposed functions listed here.
Do we want to do this to keep such file shorter?

LeoComandini:
To my understanding you should be able to produce a recoverable signature while grinding R, so, if you feel is necessary, you may want to change 05d27ae3169e513f31c714fd37278fb05cb07144 message

CHANGES.md
jgriffiths: @LeoComandini  yes, the idea here is to only list the changes that users might need to account for when upgrading, and we don't want these to get lost in the noise when we add new calls. If there are no API changes in CHANGES.md when upgrading, we want users to be able to tell that at a glance.

Users can get general info about a release from the release page, git history, generated docs, and the release announcement, all of which are better placed to detail new features etc.

jgriffiths:
> To my understanding you should be able to produce a recoverable signature while grinding R, so, if you feel is necessary, you may want to change [05d27ae](https://github.com/ElementsProject/libwally-core/commit/05d27ae3169e513f31c714fd37278fb05cb07144) message

Good point, thanks - I've updated the commonit comment as suggested :)

src/pullpush.h
```diff
@@ -0,0 +1,92 @@
...
+/* Calls pull_failed() (and returns 0) if length would exceed remaining *max */
+size_t pull_varlength(const unsigned char **cursor, size_t *max);
+
+/**
+ * Functions to parse a subfield within an area.
+ *
+ * pull_subfield_start: initializes subcursor/submax to subfield_len at
+ * cursor.  These can then be used with pull_ methods like normal.
+ *
+ * end_subfield_pull: updates *cursor and *max to point after the subfield has
```
LeoComandini, line 81: nit: s/end_subfield_pull/pull_subfield_end/

tools/build_python_cffi.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
LeoComandini, line 1: Running `./tools/build_python_cffi.sh` on my end, I spotted 2 differences:
```
$ ./tools/build_python_cffi.sh | grep wally_tx_get_total_output_satoshi
    ('wally_tx_get_total_output_satoshi', c_int, [POINTER(wally_tx), c_ulonglong_p]),
$ cat src/test/util.py | grep wally_tx_get_total_output_satoshi
    ('wally_tx_get_total_output_satoshi', c_int, [POINTER(wally_tx), c_ulonglong]),
```
```
$ ./tools/build_python_cffi.sh | grep wally_unknowns_map_find
    ('wally_unknowns_map_find', c_int, [POINTER(wally_unknowns_map), c_void_p, c_ulong, c_ulong_p]),
$ leonardo@leonardo-TECRA-A50-C:~/greenaddress/libwally-core$ cat src/test/util.py | grep wally_unknowns_map_find
```

tools/build_python_cffi.sh
```diff
@@ -0,0 +1,35 @@
...
+    grep '^int ' | \
+    sed -e "s/(/',c_int,[/g" \
+        -e "s/^int /    ('/g" \
+        -e 's/, /,/g' -e 's/const //g' | \
+    perl -pe 's/int .*?[,)]/c_int,/g' | \
+    perl -pe 's/size_t \*.*?[,)]/c_ulong_p,/g' | \
+    perl -pe 's/size_t .*?[,)]/c_ulong,/g' | \
+    perl -pe 's/uint32_t \*.*?[,)]/c_uint_p,/g' | \
+    perl -pe 's/uint32_t .*?[,)]/c_uint,/g' | \
+    perl -pe 's/uint64_t \*.*?,/POINTER(c_ulonglong),/g' | \
```
LeoComandini, line 23: Why do we need `POINTER(c_ulonglong)` when we have `c_ulonglong_p`?

src/pullpush.h
jgriffiths: Cheers, updated!

tools/build_python_cffi.sh
jgriffiths: Ah we don't, thats a missed leftover from converting to allow int64 funcs to return values directly, I'll fix in place.

tools/build_python_cffi.sh
jgriffiths: Ive updated the wrapper generator again and re-generated if you could review again, thanks!

include/wally_psbt.h
```diff
@@ -122,338 +181,399 @@ struct wally_psbt {
...
- * Set the witness_utxo in an input
+ * Set the witness_utxo in an input.
  *
  * :param input: The input to update.
  * :param witness_utxo: The witness utxo for this input if it exists.
  */
 WALLY_CORE_API int wally_psbt_input_set_witness_utxo(
     struct wally_psbt_input *input,
-    struct wally_tx_output *witness_utxo);
+    const struct wally_tx_output *witness_utxo);
```
LeoComandini, line 398: Should we mention in `CHANGES.md` that some API parameters are now const?

include/wally_psbt.h
```diff
@@ -463,121 +583,467 @@ WALLY_CORE_API int wally_psbt_init_alloc(
...
+    size_t len,
     struct wally_psbt **output);
 
 /**
- * Get length of psbt when serialized to bytes.
+ * Get the length of a PSBT when serialized to bytes.
  *
  * :param psbt: the PSBT.
- * :param len: Length in bytes when serialized.
+ * :param written: Destination for the length in bytes when serialized.
```
LeoComandini, line 624: `flags` description is missing. Also for below functions

include/wally_psbt.h
jgriffiths: Moving from non-const to const isn't an API change for the caller, although the other way around would be. So there's no need to mention this here I think.

include/wally_psbt.h
jgriffiths: Well spotted, thanks! I will update them in a new patch, thanks.

jgriffiths: Done.

src/psbt.c
```diff
@@ -1,44 +1,72 @@
...
-int wally_keypath_map_init_alloc(size_t alloc_len, struct wally_keypath_map **output)
+static bool is_equal_pubkey(const unsigned char *l, size_t l_len, const unsigned char *r, size_t r_len)
+{
+    if (!l || !r)
+        return l == r; /* Equal if both NULL only */
+
+    return l_len == r_len && memcmp(l, r, l_len) == 0;
+}
+
+int wally_keypath_map_init_alloc(size_t allocation_len, struct wally_keypath_map **output)
```
LeoComandini, line 54: For keypaths, partial sigs and unknowns maps we have some functions (init alloc, free, find, add, replace) which are almost identical. Do you think it's worth to abstract them somehow and share some code?

jgriffiths: Yes, absolutely!

The reality is that the initially implemented PSBT interface isn't fit for purpose, and will change more beyond the changes here. As you point out for example, partial sigs and unknowns maps are exactly the same (key->value maps) and keypath could be implemented using the same code if the keypath is stored as it is in the PSBT file (key->value with the fingerprint as the first element in the path). That would allow us to share the code for all 3 maps and get rid of 2 of the types and associated wrapping code (and fix the naming inconsistencies and plural confusion in the type/function names).

Also, PSBT is the 2nd library addition that requires structure manipulation and so some of the code that supports this in transaction.c should be generalised so it can be reused when we add the 3rd (descriptors or whatever).

However - These changes will break the PSBT interface completely. My view is that no one can have been using it previously due to the bugs, as evidenced by c-lightning having to rewrite and fix large parts of the code in order to get things working (and they still have a bunch of workarounds in their tree after moving to this branch).

So, once this PR is merged I want to push the bigger interface changes before the next release and just document the change in CHANGES.md as 'the PSBT interface has changed and can now be considered final'. That deserves its own PR though so any bike-shedding over the interface can happen there.

src/psbt.c
```diff
@@ -287,303 +332,301 @@ int wally_unknowns_map_free(struct wally_unknowns_map *unknowns)
...
 
-int wally_psbt_input_set_partial_sigs(
-    struct wally_psbt_input *input,
-    struct wally_partial_sigs_map *partial_sigs)
+int wally_psbt_elements_input_set_value_blinder(struct wally_psbt_input *input,
+                                                const unsigned char *value_blinder,
+                                                size_t value_blinder_len)
 {
-    struct wally_partial_sigs_map *result_partial_sigs;
+    if (!input)
```
LeoComandini, line 556: I think we should check (here and below) the length of the value set against the values in https://github.com/ElementsProject/libwally-core/blob/master/include/wally_elements.h#L12. This would be consistent with what is done in `wally_psbt_elements_output_set_blinding_pubkey`

jgriffiths: Yes, agreed. I would have added this but at present the naming of things leaves me a little confused as to what these members actually are. Is "asset_blinder" _actually_ an asset blinding factor? If so it  should be called abf and passed with abf_len so SWIG picks it up automatically and the docs are easier to follow and maintain.

In general the code prior to this patch series does no testing of values, for example pubkey validity and preventing dupes in KV maps. I think tightening up the validation will happen iteratively after this patch series is applied.

I will add a FIXME to check these values until I can confirm what they actually represent by following through to where they are used. These setters will probably be replaced with macro versions at some point as well.

jgriffiths: FIXME added.

src/psbt.c
```diff
@@ -712,825 +801,1074 @@ int wally_psbt_init_alloc(
...
+
+    if (ret != WALLY_OK ||
+        (inputs_allocation_len && !result->inputs) || (outputs_allocation_len && !result->outputs)) {
+        wally_psbt_free(result);
+        return ret != WALLY_OK ? ret : WALLY_ENOMEM;
     }
 
-    result->inputs = new_inputs;
-    result->num_inputs = 0;
+    /* Version is always 0 */
```
LeoComandini, line 828: `wally_tx_init_alloc` allows to set the transaction version, should we allow the user to set the psbt version too in `wally_psbt_init_alloc` and its elements variant? It could be needed in the future.

src/psbt.c
```diff
@@ -1541,648 +1879,508 @@ int wally_psbt_from_bytes(
...
-        psbt_input_get_length(input, &input_len);
-        out += input_len;
-    }
-    for (i = 0; i < psbt->num_outputs; ++i) {
-        struct wally_psbt_output *output = &psbt->outputs[i];
-        size_t output_len;
-        psbt_output_get_length(output, &output_len);
-        out += output_len;
-    }
+    origin_len = 4;     /* Start with 4 bytes for fingerprint */
```
LeoComandini, line 1993: nit: s/4/BIP32_KEY_FINGERPRINT_LEN/ (also few lines below)

src/psbt.c
jgriffiths: Yes, good idea - pushed.

src/psbt.c
jgriffiths: Fixed, thanks!

src/psbt.c
```diff
@@ -2192,347 +2388,236 @@ int wally_psbt_to_base64(
...
-fail:
-    wally_psbt_free(result);
     return ret;
 }
 
-int wally_sign_psbt(
-    struct wally_psbt *psbt,
-    const unsigned char *key,
-    size_t key_len)
+int wally_psbt_sign(struct wally_psbt *psbt,
```
LeoComandini, line 2600: I wonder if we should add a `flags` parameter here. Currently it signs with `EC_FLAG_ECDSA | EC_FLAG_GRIND_R`, but users may want to use different flags. OTOH if `EC_FLAG_SCHNORR` should be used it can probably be deduced from the signed scripts and `EC_FLAG_RECOVERABLE` should not be accepted, so it may be ok as it is.

jgriffiths: I assume we can tell ECDSA from Schnorr by the witness/scripts, but we should probably accept at least the grinding flag (super low power devices may opt not to grind if signing takes too long, for example). And there may be other relevant flags in the future like alternate nonce functions or something relating to signature aggregation. I need to think on this a bit.

However this PR hasn't addressed the signing code yet, that will probably be a separate PR, looks to me like the current impl can be fairly easily made to crash. This PR can't fix everything that needs doing though, or it will end up well over its current ~200 commits :(

LeoComandini:
I went through the changed code and it looks good to me.
As said in the above comments other PRs will follow to cover some issues, add the missing tests and finalize the psbt interface.
So IMHO we should be good in merging this PR at this stage.

utACK 1b616660718aa02d72769999d7a69b65d9dec2a9

src/psbt.c
jgriffiths: I've added the ability to control grinding (and extend with other flags in the future).

LeoComandini:
New changes looks good too,

utACK c9c45b68c5a18ca1b5bca41f3c3c8350a23b1bd8

PR #211, opened 2020-07-28
PSBT updates, continued

jgriffiths:
Further updates to PSBT related code motivated by c-lightning usage.

This should be the last of the major interface changes although we might need some small tweaks when we get a second PSBT user.

@achow101 If you could review the final versions of include/wally_psbt.h and src/psbt.c that would be useful. Also the coverage report shows that finalize_p2pkh is never called, so we could use a test case for that added to src/data/psbt.json if you have one handy.


src/swig_python/swig.i
```diff
@@ -354,9 +324,12 @@ static void destroy_words(PyObject *obj) { (void)obj; }
...
 %rename("bip32_key_with_tweak_from_parent_path") bip32_key_with_tweak_from_parent_path_alloc;
-%rename("keypath_map_init") wally_keypath_map_init_alloc;
-%rename("partial_sigs_map_init") wally_partial_sigs_map_init_alloc;
+%rename("psbt_clone") wally_psbt_clone_alloc;
 %rename("psbt_get_global_tx") wally_psbt_get_global_tx_alloc;
+%rename("psbt_get_input_final_witness") wally_psbt_get_input_final_witness_alloc;
+%rename("psbt_get_input_utxo") wally_psbt_get_input_utxo_alloc;
+%rename("psbt_get_input_pegin_tx") wally_psbt_get_input_pegin_tx_alloc;
+%rename("psbt_get_input_witness_utxo") wally_psbt_get_input_witness_utxo_alloc;
 %rename("psbt_init") wally_psbt_init_alloc;
```
LeoComandini, line 333: We should the same for `wally_psbt_elements_init_alloc`

src/psbt.c
```diff
@@ -2333,406 +2092,370 @@ int wally_psbt_sign(struct wally_psbt *psbt,
...
+
+    return input->redeem_script ? finalize_p2sh_wrapped(input) : true;
+}
+
+static bool finalize_multisig(struct wally_psbt_input *input,
+                              const unsigned char *out_script, size_t out_script_len,
+                              bool is_witness, bool is_p2sh)
+{
+    unsigned char sigs[EC_SIGNATURE_LEN * 15];
+    uint32_t sighashes[EC_SIGNATURE_LEN * 15];
```
LeoComandini, line 2294: s/EC_SIGNATURE_LEN * 15/15/

src/test/util.py
```diff
@@ -308,57 +287,65 @@ class wally_psbt(Structure):
...
+    ('wally_psbt_input_is_finalized', c_int, [POINTER(wally_psbt_input), c_ulong_p]),
     ('wally_psbt_input_set_abf', c_int, [POINTER(wally_psbt_input), c_void_p, c_ulong]),
     ('wally_psbt_input_set_asset', c_int, [POINTER(wally_psbt_input), c_void_p, c_ulong]),
     ('wally_psbt_input_set_claim_script', c_int, [POINTER(wally_psbt_input), c_void_p, c_ulong]),
+    ('wally_psbt_input_set_final_scriptsig', c_int, [POINTER(wally_psbt_input), c_void_p, c_ulong]),
+    ('wally_psbt_input_set_final_witness', c_int, [POINTER(wally_psbt_input), POINTER(wally_tx_witness_stack)]),
     ('wally_psbt_input_set_genesis_blockhash', c_int, [POINTER(wally_psbt_input), c_void_p, c_ulong]),
-    ('wally_psbt_input_set_peg_in_tx', c_int, [POINTER(wally_psbt_input), POINTER(wally_tx)]),
+    ('wally_psbt_input_set_keypaths', c_int, [POINTER(wally_psbt_input), POINTER(wally_map)]),
+    ('wally_psbt_input_set_utxo', c_int, [POINTER(wally_psbt_input), POINTER(wally_tx)]),
```
LeoComandini, line 319: Lines generated with `./tools/build_python_cffi.sh` are in a different order w.r.t the ones in `src/test/util.py`. We may want to have them consistent.

src/psbt.c
jgriffiths: good catch, thanks!!

src/psbt.c
```diff
@@ -2099,103 +1795,45 @@ int wally_psbt_to_base64(const struct wally_psbt *psbt, uint32_t flags, char **o
...
+    clear_and_free(result, b64_safe_len);
+    clear_and_free(buff, len);
     return ret;
 }
 
-#define COMBINE_BYTES(typ, member) \
-    if (!dst->member && src->member && \
-        (ret = wally_psbt_ ## typ ## _set_ ## member(dst, src->member, src->member ## _len)) != WALLY_OK) \
-        return ret
+#define COMBINE_BYTES(typ, member)  do { \
```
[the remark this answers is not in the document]
jgriffiths: The do/while loop just allows us to add a colon at the end of the macro invocation which reads a bit better when there are multiple calls to the same macro one after the other. When the macros in transaction.c and psbt are generalised in the future I will use this form I suspect.

The second check is redundant (and the compiler will eliminate it), but i want to keep it there to point out that currently present-but-empty is indicated by having a pointer but 0 length which is counter to the rest of wallys view on how varbuffs should be held. In a future patchset I'd like to replace these with a bitflag indicating what members are present, but that will have to wait as I'm out of time for now.

src/swig_python/swig.i
jgriffiths: Done, thanks!

src/test/util.py
jgriffiths: Yes, the ones in the file are sorted by vim :%sort which differs from /usr/bin/sort.

The current sed/perl generation approach is not sufficiently general to be used for generating wally.hpp, or the swig python/java required typemaps, all of which are constantly out of date as was the cffi list here. So my plan is to write a python parser which can generate all of these and have it write the new values in place. That will resolve the issue of sorting since there will be no manual changes to this list following that. Until then I'm OK with updating this manually and living with the sort difference.

src/psbt.c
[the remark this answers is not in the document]
jgriffiths: I've raised https://github.com/ElementsProject/libwally-core/issues/212 for this.

src/test/util.py
jgriffiths: I've raised https://github.com/ElementsProject/libwally-core/issues/212 for this.

src/psbt.c
```diff
@@ -2333,406 +2092,370 @@ int wally_psbt_sign(struct wally_psbt *psbt,
...
-            unsigned char txid[WALLY_TXHASH_LEN];
+        sighash = input->sighash ? input->sighash : WALLY_SIGHASH_ALL;
 
-            if ((ret = wally_tx_get_txid(input->non_witness_utxo, txid, sizeof(txid))) != WALLY_OK) {
-                return ret;
-            }
-            if (memcmp(txid, txin->txhash, sizeof(txid)) != 0) {
-                return WALLY_EINVAL;
-            }
+        if (input->utxo) {
```
LeoComandini, line 2146: Due to recent change in to hardware wallets, the full previous transaction is provided for segwit inputs https://github.com/bitcoin/bitcoin/pull/19215.

This means that we compute and sign the pre-segwit signature hash for segwit inputs, see https://github.com/ElementsProject/libwally-core/issues/213, https://github.com/bitcoin/bitcoin/issues/19650.

jgriffiths: Yes, this behaviour was buggy in the old code and remains so in the cleaned up code, I think this bug needs to be addressed in a new PR rather than getting caught up with the interface changes and general cleanup in this one.

Since we have https://github.com/ElementsProject/libwally-core/issues/213 to cover this issue I'll fix that bug separetly with a candidate PR we can hopefully get @Fonta1n3 to help us test.

Fonta1n3: Absolutely, you guys ever need anything at all just tag me and I will get on it. Really appreciate all of that you do and would be happy to test help in anyway I can.

src/swig_python/python_extra.py_in
```diff
@@ -162,6 +162,16 @@ wif_to_public_key = _wrap_bin(wif_to_public_key, _wif_to_public_key_len, resize=
...
+psbt_get_input_redeem_script = _wrap_bin(psbt_get_input_redeem_script, psbt_get_input_redeem_script_len, resize=True)
+psbt_get_input_witness_script = _wrap_bin(psbt_get_input_witness_script, psbt_get_input_witness_script_len, resize=True)
+psbt_get_input_final_scriptsig = _wrap_bin(psbt_get_input_final_scriptsig, psbt_get_input_final_scriptsig_len, resize=True)
+psbt_get_input_signature = _wrap_bin(psbt_get_input_signature, psbt_get_input_signature_len, resize=True)
+psbt_get_input_unknown = _wrap_bin(psbt_get_input_unknown, psbt_get_input_unknown_len, resize=True)
+
+psbt_get_output_redeem_script = _wrap_bin(psbt_get_output_redeem_script, psbt_get_output_redeem_script_len, resize=True)
+psbt_get_output_witness_script = _wrap_bin(psbt_get_output_witness_script, psbt_get_output_witness_script_len, resize=True)
+psbt_get_output_unknown = _wrap_bin(psbt_get_output_unknown, psbt_get_output_unknown_len, resize=True)
+
```
LeoComandini, line 174: also `psbt_to_bytes` should be wrapped

include/wally_psbt.h
```diff
@@ -535,12 +547,62 @@ WALLY_CORE_API int wally_psbt_set_global_tx(
...
+ */
+WALLY_CORE_API int wally_psbt_remove_output(
+    struct wally_psbt *psbt,
+    uint32_t index);
+
 /**
  * Create a PSBT from its serialized bytes.
  *
  * :param bytes: Bytes to create the PSBT from.
  * :param len: Length of ``bytes`` in bytes.
```
LeoComandini, line 602: s/len/bytes_len/ so swig handle this correctly

LeoComandini:
ACK 633dac83c926dd8c337893295e458a6ea279a536, code review, tested part of the python bindings.

src/swig_python/python_extra.py_in
jgriffiths: Fixed, thanks!

include/wally_psbt.h
jgriffiths: Updated, thanks!

PR #215, opened 2020-08-08
More misc updates

jgriffiths:
Fixes for elements PSBT comparing and broken signing due to cores v20.1 PSBT changes.


src/test/test_psbt.py
```diff
@@ -113,5 +113,72 @@ def test_serialization(self):
...
+
+        # Check invalid args
+        self.assertEqual(wally_map_init_alloc(0, None), WALLY_EINVAL)
+        self.assertEqual(wally_map_init_alloc(0, m), WALLY_OK)
+
+        for args in [(None, key1, key1_len, val,  val_len), # Null map
+                     (None, None, key1_len, val,  val_len), # Null key
+                     (None, key1, 0,        val,  val_len), # 0 length key
+                     (None, key1, key1_len, None, val_len), # Null value
+                     (None, key1, key1_len, val,  0)]:      # 0 length value
```
LeoComandini, line 133:
```suggestion
                     (m,    None, key1_len, val,  val_len), # Null key
                     (m,    key1, 0,        val,  val_len), # 0 length key
                     (m,    key1, key1_len, None, val_len), # Null value
                     (m,    key1, key1_len, val,  0)]:      # 0 length value
```

jgriffiths: Good catch! Updated, thanks!

LeoComandini:
ACK b8d7ea91049c3d5522768c77c8bfe4936cbabbd7, code review, tested wally_map_sort through python bindings, checked that strings in the new test case match the ones in the issue.

Github CI failures looks unrelated and definitely not caused by the last pushed changes.

PR #216, opened 2020-08-13
script: add miniscript-compatible 2of2 CSV scriptpubkey support

jgriffiths:
This script is 3 bytes shorter for the common (non-recovery) case.


src/swig_python/python_extra.py_in
```diff
@@ -102,9 +102,13 @@ def _script_push_from_bytes_len_fn(data, flags):
             break
     return push_len + opcode_len
 script_push_from_bytes = _wrap_bin(script_push_from_bytes, _script_push_from_bytes_len_fn, resize=True)
-def _spkcsv2of2_1_len_fn(sigs, csv_blocks, flags):
+def _spkcsv2of2_1_opt_len_fn(sigs, csv_blocks, flags):
```
LeoComandini, line 105: nit: s/sigs/pubkeys/ (also below)

however here it does not really matter so to me it's fine to leave it as it is

src/test/test_script.py
```diff
@@ -200,8 +200,9 @@ def test_scriptpubkey_csv_2of2_then_1_from_bytes(self):
             (MPK_2, MPK_2_LEN, 1, 0, None, out_len), # Null output
         ]
         for args in invalid_args:
-            ret = wally_scriptpubkey_csv_2of2_then_1_from_bytes(*args)
-            self.assertEqual(ret, (WALLY_EINVAL, 0))
+            for fn in [wally_scriptpubkey_csv_2of2_then_1_from_bytes,
+                       wally_scriptpubkey_csv_2of2_then_1_from_bytes_opt]:
+                self.assertEqual(fn(*args), (WALLY_EINVAL, 0))
 
         # Valid cases
```
LeoComandini, line 207: Should we add also a valid test case? It may not be worth it though, it does not add much value.

src/script.c
```diff
@@ -712,6 +712,53 @@ int wally_scriptpubkey_csv_2of2_then_1_from_bytes(
...
+    memcpy(bytes_out, bytes + EC_PUBLIC_KEY_LEN, EC_PUBLIC_KEY_LEN);
+    bytes_out += EC_PUBLIC_KEY_LEN;
+    *bytes_out++ = OP_CHECKSIGVERIFY;
+    *bytes_out++ = EC_PUBLIC_KEY_LEN;
+    memcpy(bytes_out, bytes, EC_PUBLIC_KEY_LEN);
+    bytes_out += EC_PUBLIC_KEY_LEN;
+    *bytes_out++ = OP_CHECKSIG;
+    *bytes_out++ = OP_IFDUP;
+    *bytes_out++ = OP_NOTIF;
+    *bytes_out++ = csv_len & 0xff;
```
LeoComandini, line 753: If csv blocks is in the interval 1-16, Core rejects the recovery path.

With `testmempoolaccept` I obtained:

    "reject-reason": "non-mandatory-script-verify-flag (Data push larger than necessary)"

This error did not happen if I was spending with the other path.

My guess is that for N in 1-16, we should replace `<1> <n>` with `OP_N`, so for instance:

    >>> wally.scriptpubkey_csv_2of2_then_1_from_bytes_opt(b'\x00'*66, 7, 0).hex()
    '21000000000000000000000000000000000000000000000000000000000000000000ad21000000000000000000000000000000000000000000000000000000000000000000ac73640107b268'
    
should be

    '21000000000000000000000000000000000000000000000000000000000000000000ad21000000000000000000000000000000000000000000000000000000000000000000ac736457b268'

This affects `wally_scriptpubkey_csv_2of2_then_1_from_bytes` and `wally_scriptpubkey_csv_2of3_then_2_from_bytes` too.
Should we fix them so that they produce a script accepted by the mempool too? That may break existing wallets that may not be able to reconstruct past scripts. OTOH, using less than 16 csv blocks seems quite odd and definitely not realistic for meaningful settings, so it may be ok to change it now. Another option is to fail in both cases if csv blocks is 16 or below, which to me is acceptable.

src/swig_python/python_extra.py_in
jgriffiths: This was wrong in a couple of places so I updated them all, thanks!

src/test/test_script.py
jgriffiths: Yes, I will add valid cases once we decide if we will expose this variant.

src/script.c
jgriffiths: Good catch! I think we should fail in this case, its almost certainly going to be an error for a user to set a recovery time this low.

LeoComandini:
ACK 448255e7010e67842814d27f3c756dd6c0265b2f, code review, tested `wally_scriptpubkey_csv_2of2_then_1_from_bytes_opt` in a local regtest spending with both paths.

jgriffiths:
Merged with test, closing.

PR #217, opened 2020-08-13
[wip] Implement redesigned PSET

achow101:
PSET is being redesigned in order to actually do everything that elements transactions can do while also preserving privacy.

See https://github.com/ElementsProject/elements/pull/900 for more details about this, including a specification for the new PSET.

### DIFF
--- a/src/psbt.c
+++ b/src/psbt.c
@@ -48,23 +48,27 @@
 #define PSBT_OUT_BIP32_DERIVATION 0x02
 
 #ifdef BUILD_ELEMENTS
-#define PSET_IN_VALUE 0x00
-#define PSET_IN_VALUE_BLINDER 0x01
-#define PSET_IN_ASSET 0x02
-#define PSET_IN_ASSET_BLINDER 0x03
-#define PSET_IN_PEG_IN_TX 0x04
-#define PSET_IN_TXOUT_PROOF 0x05
-#define PSET_IN_GENESIS_HASH 0x06
-#define PSET_IN_CLAIM_SCRIPT 0x07
+#define PSET_GLOBAL_SCALRA 0x00
 
-#define PSET_OUT_VALUE_COMMITMENT 0x00
-#define PSET_OUT_VALUE_BLINDER 0x01
-#define PSET_OUT_ASSET_COMMITMENT 0x02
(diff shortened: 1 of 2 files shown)


jgriffiths:
@niftynei Heads up that the elements interface will be changing per the linked PR.

Hi @achow101, for the presence/absence of fields in inputs and outputs I am leaning towards using a bitset and removing the hacks to malloc(1) buffers leaving their length 0. Can you comment as to how likely the Elements interface is to change at this time?

achow101:
> Can you comment as to how likely the Elements interface is to change at this time?

I'm not sure how it will change at this time, there still some work being done on the design of the Blinder role.

jgriffiths:
> I'm not sure how it will change at this time, there still some work being done on the design of the Blinder role.

OK, I imagine we want the updated Elements support in the next release, since we can then get all the PSBT interface changes done within one release.

src/psbt.c
```diff
@@ -48,23 +48,27 @@
 #define PSBT_OUT_BIP32_DERIVATION 0x02
 
 #ifdef BUILD_ELEMENTS
-#define PSET_IN_VALUE 0x00
-#define PSET_IN_VALUE_BLINDER 0x01
-#define PSET_IN_ASSET 0x02
-#define PSET_IN_ASSET_BLINDER 0x03
+#define PSET_GLOBAL_SCALRA 0x00
```
niftynei, line 51:
```suggestion
#define PSET_GLOBAL_SCALAR 0x00
```
extremely drive by nit lol

jgriffiths:
@achow I have started implementing these changes locally including tests, swig/C++ wrappers and code de-duplication.

I'm looking at the malloc hack in src/psbt.c now and while I can see it lets you differentiate between "field not present" and "field present but null", I'm wondering under what circumstances this is actually required/legal? The test cases you added in commit b5cde6e16a000a345dd7ffd8977872f8e0e33e4e all fail to parse using cores decodepsbt rpc command due to duplicate keys. Is there any reason I shouldn't rip this code out?

jgriffiths:
Closing this in favour of https://github.com/ElementsProject/libwally-core/pull/218 which implements the PSET BIP as currently specified. Please update there when the spec changes/is finalised.

PR #218, opened 2020-08-17
Elements psbt updates

jgriffiths:
Updates the Elements PSBT support to match the latest changes as per ElementsProject/elements#900.

Since that spec is not finalised - yet more changes may be required.


jgriffiths:
Superseded with now merged PSBT/PSETv2 support, closing.

PR #221, opened 2020-08-28
bip32: Don't call privkey_tweak_add by empty tweakSum

k-matsuzawa:
There was a problem when we confirmed the operation in combination with the latest secp256k1-zkp.
(does not occur problem with secp256k1-zkp currently included in libwally-core)

Cause:
In the latest secp256k1, logic to check invalid privkey with privkey_tweak_add has been added.
ext_key::pub_key_tweak_sum is all-zero on initialized.
Therefore, the error occurred in bip32_key_from_parent().

### DIFF
--- a/src/bip32.c
+++ b/src/bip32.c
@@ -337,6 +337,16 @@ int bip32_key_unserialize_alloc(const unsigned char *bytes, size_t bytes_len,
     return ret;
 }
 
+static int bip32_privkey_tweak_add(const secp256k1_context *ctx, unsigned char *seckey,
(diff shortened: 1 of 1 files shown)


jgriffiths:
Hi @k-matsuzawa can you verify that https://github.com/ElementsProject/libwally-core/pull/222 works as intended, and if so I'll merge that version and close this.

Thanks!

k-matsuzawa:
@jgriffiths   Thank you! I could verify #222 works as intended. :tada:

jgriffiths:
@k-matsuzawa Thanks, I have merged it and will close this one :)

PR #247, opened 2020-11-20
Make emscripten emcc options configurable

shesek:
This also changes the default optimization level to 3 (tested to work) and renames the target directory to `wasm_dist`.

### DIFF
--- a/tools/build_wasm.sh
+++ b/tools/build_wasm.sh
@@ -10,5 +10,16 @@ $PWD/tools/cleanup.sh && $PWD/tools/autogen.sh
 export CFLAGS="-fno-stack-protector"
 emconfigure ./configure --build=$HOST_OS ac_cv_c_bigendian=no --disable-swig-python --disable-swig-java --enable-export-all --enable-elements --disable-ecmult-static-precomputation
 emmake make -j $num_jobs
-#FIXME: this is just an example, numerous functions are missing
-emcc -O2 -s "EXTRA_EXPORTED_RUNTIME_METHODS=['getValue', 'UTF8ToString']" -s "EXPORTED_FUNCTIONS=['_wally_init','_wally_get_secp_context','_wally_secp_randomize', '_wally_free_string', '_bip39_get_wordlist', '_bip39_mnemonic_from_bytes', '_bip39_mnemonic_to_seed']" ./src/.libs/*.o src/secp256k1/src/*.o src/ccan/ccan/crypto/*/.libs/*.o ./src/ccan/ccan/str/hex/.libs/*.o -o wallycore.html --shell-file contrib/shell_minimal.html
+
+: ${OPTIMIZATION_LEVEL:=3}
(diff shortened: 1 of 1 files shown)


shesek:
Wasn't sure which functions should get exported by default, should it be all of them?

jgriffiths:
> Wasn't sure which functions should get exported by default, should it be all of them?
Hi @shesek, Ideally we should export all external functions, yes. A couple of comments:

- Elements support should be optional, controlled with an argument to the script like '--enable-elements'
- `--enable-export-all` should be removed from the configure arguments
- If `emcc` can take an file to list the exports rather than a command line option, it would be good to add support for generating this file via `./tools/build_wrappers.py`, alternately the script itself can be updated by `build_wrappers.py` in the same way as the other generated code (this ensures it will stay up to date going forward when new functions are added).
- It might be cleaner to pull your own branch, cherry pick Lawrences commit, add your changes and PR that, as this branch needs rebasing.
- I will see about updating our greenaddrress build images to include the wasm deps so this can me merged for the next release.

We may want to put something in README.md on safety as well, I don't know if the constant time code in secp or the AES code will remain constant time once transpiled. So users may want to avoid signing or encrypting using this build.

Cheers!

jgriffiths:
@shesek Might be a plan to hold off for a day or two, we will merge the changes from https://github.com/ElementsProject/libwally-core/pull/248/ soon which should give us an up to date docker image base to build masm support from

shesek:
Thanks for your review and help!

> If `emcc` can take an file to list the exports rather than a command line option, it would be good to add support for generating this file via `./tools/build_wrappers.py`, ...

The `build_wasm.sh` script could get extract the list of functions in directly from the headers to create the json array, with something like this:

`gcc -E include/*.h | sort | uniq | grep '^ *int ' | grep '(' | sed -e 's/^ *int //g' -e 's/(.*//g' | egrep '^wally_|^bip' | jq -R -s -c '[split("\n")[]|select(length>0)] | map("_"+.)'`

(Based on the code from `build_wrappers.py`, with `jq` to serialize as an array. Can we assume to have jq`` available?)

> the `client/src/lib/libwally.js` wrappers in the esplora commit look like they might be useful to provide with wally itself, particularly if they can be auto-generated. 

Yes, definitely, auto-generating wrappers to expose whole of libwally would be great. I only needed two so it was easier to just implement them manually, but I'll look into the tooling available for auto generation.

jgriffiths:
@shesek, I'm implementing autogeneration now, please hold on for a bit and I'll push the PR after testing...

jgriffiths:
@shesek please see https://github.com/ElementsProject/libwally-core/pull/249 which auto-generates the function lists, adds `--enable-elements` to optionally include elements support, and lets the user override the exported functions by setting `EXPORTED_FUNCTIONS` themselves.

Our CI has passed that PR, so we should be merging it shortly (after some manual testing). Please feel free to open a new PR or issue if there's something broken or anything further you need.

shesek:
@jgriffiths Looks great! :+1:

PR #249, opened 2020-11-25
WASM: auto-export library functions, make elements support optional

jgriffiths:


shesek:
It seems like the wasm bundle I'm [building for esplora](https://github.com/Blockstream/esplora/pull/276/files#diff-bffdc20d769fb82fa249b9cf47c3f8086771e1b9b4043a9fd44ba61e5a063005R1) grew from 79kb to 133kb, could this be due `--enable-export-all`? Why was it added?

jgriffiths:
> Edit: Nope, it is not that, I tried without `--enable-export-all` and got the same size. Maybe libwally just got bigger since lawrence's original `webassembly` branch?

--enable-export-all just disables symbol visibility. I think the size is controlled by which functions you choose to export. If you pass your own list of just the functions you need, the optimiser will eliminate all the dead code leaving you with a minimal wasm file.

PR #271, opened 2021-04-03
add Dockerfile

secinthenet:


secinthenet:
Quick question: even though I'm exporting `CC="clang-10" CXX="clang++-10"`, it seems that the tests run with gcc, which is presumably why the clear test fails and I need to disable it?

jgriffiths:
> Quick question: even though I'm exporting `CC="clang-10" CXX="clang++-10"`, it seems that the tests run with gcc, which is presumably why the clear test fails and I need to disable it?

The tests should be built with the same compiler as the library itself, I'd need to try this docker build to see whats going on. However I have noticed on debian that using the named compiler may not work sometimes, I have had more success (with e.g. gdk) installing clang versions into unique directories and using PATH to control the compiler used.

Thanks for this contribution! it may be a little while before we review and merge it but rest assured we will take a look as soon as we get time.

PR #281, opened 2021-05-27
Add Bech32m implementation and extend test case

fadedreamz:
This commit adds bech32m witaddr implementation for
non zero witver addresses. Test cases have been updated
to reflect witver > 0 addresses.

Signed-off-by: Mohammad Nazmul Alam <fadedreamz@gmail.com>


src/bech32.c
```diff
@@ -21,36 +21,44 @@
...
+        (-((b >> 2) & 1) & 0x1ea119faUL) ^
+        (-((b >> 3) & 1) & 0x3d4233ddUL) ^
+        (-((b >> 4) & 1) & 0x2a1462b3UL);
 }
 
-static const char *charset = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+static uint32_t bech32_final_constant(bech32_encoding enc) {
+    if (enc == BECH32_ENCODING_BECH32) return 1;
+    if (enc == BECH32_ENCODING_BECH32M) return 0x2bc830a3;
+    assert(0);
```
jgriffiths, line 45: we can't use assert for error handling in wally, only return error codes to the caller.

src/bech32.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _BECH32_H_
+#define _BECH32_H_
+
+typedef enum {
+    BECH32_ENCODING_NONE,
+    BECH32_ENCODING_BECH32,
+    BECH32_ENCODING_BECH32M
+} bech32_encoding;
+
+#endif
```
jgriffiths, line 10: constants used only in one file shouldn't be placed in a header.

src/test/test_bech32.py
```diff
@@ -56,7 +73,7 @@ def test_segwit_address(self):
             # Encode the script and make sure the address matches
             script_buf, script_len = make_cbuffer(script_hex)
             ret, retstr = wally_addr_segwit_from_bytes(script_buf, script_len, utf8(family), 0)
-            if script_ver != 0:
+            if script_ver > 16:
```
jgriffiths, line 76: we don't want to accept versions that we don't know how to parse/verify in wally, see 5c770b446db0bf76aebfe3293ca4e64329beaa21

src/bech32.c
```diff
@@ -21,36 +21,44 @@
...
-           (-((b >> 0) & 1) & 0x3b6a57b2UL) ^
-           (-((b >> 1) & 1) & 0x26508e6dUL) ^
-           (-((b >> 2) & 1) & 0x1ea119faUL) ^
-           (-((b >> 3) & 1) & 0x3d4233ddUL) ^
-           (-((b >> 4) & 1) & 0x2a1462b3UL);
+        (-((b >> 0) & 1) & 0x3b6a57b2UL) ^
+        (-((b >> 1) & 1) & 0x26508e6dUL) ^
+        (-((b >> 2) & 1) & 0x1ea119faUL) ^
+        (-((b >> 3) & 1) & 0x3d4233ddUL) ^
+        (-((b >> 4) & 1) & 0x2a1462b3UL);
```
jgriffiths, line 39: please remove whitespace changes from PR diffs

src/bech32.c
```diff
@@ -21,36 +21,44 @@
 #include "internal.h"
 #include <stdlib.h>
 #include <stdint.h>
+#include <assert.h>
```
jgriffiths, line 24: see below re: assert

src/bech32.c
```diff
@@ -21,36 +21,44 @@
...
-    1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
     -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
-    1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
 };
 
-static int bech32_encode(char *output, const char *hrp, const uint8_t *data, size_t data_len, size_t max_input_len) {
+int bech32_encode(char *output, const char *hrp, const uint8_t *data, size_t data_len, bech32_encoding enc) {
     uint32_t chk = 1;
```
jgriffiths, line 62: we use max_input_len to allow longer encoded values to use by lightning (see the git history for this file)

src/bech32.c
```diff
@@ -165,36 +179,47 @@ static int convert_bits(uint8_t *out, size_t *outlen, int outbits, const uint8_t
...
-    if (strncmp(hrp, hrp_actual, 84) != 0) goto fail;
-    if (data[0] > 16) goto fail;
-    *witdata_len = 0;
-    if (!convert_bits(witdata, witdata_len, 8, data + 1, data_len - 1, 5, 0)) goto fail;
-    if (*witdata_len < 2 || *witdata_len > 40) goto fail;
-    if (data[0] == 0 && *witdata_len != 20 && *witdata_len != 32) goto fail;
-    *witver = data[0];
-    return 1;
-fail:
+    do {
```
jgriffiths, line 207: goto is not evil, please do not reformulate control structures without need, it complicates the review process.

src/test/test_bech32.py
```diff
@@ -22,11 +26,24 @@
...
+    ["bc", 'bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7v8n0nx0muaewav253zgeav'], # Invalid program length (41 bytes)
     ["bc", 'bc10w508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7kw5rljs90'], # Invalid program length
     ["bc", 'BC1QR508D6QEJXTDG4Y5R3ZARVARYV98GJ9P'], # Invalid program length for witness version 0 (per BIP141)
     ["tb", 'tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sL5k7'], # Mixed case
     ["bc", 'bc1zw508d6qejxtdg4y5r3zarvaryvqyzf3du'], # zero padding of more than 4 bits
+    ["tb", 'tb1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vq47Zagq'], # Mixed case
+    ["bc", 'bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7v07qwwzcrf'], # more than 4 bit padding
     ["tb", 'tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3pjxtptv'], # Non-zero padding in 8-to-5 conversion
+    ["tb", 'tb1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vpggkg4j'], # Non-zero padding in 8-to-5 conversion
     ["bc", 'bc1gmk9yu'], # Empty data section
```
jgriffiths, line 47: nice tests!

jgriffiths (commented):
Hi @fadedreamz thanks for submitting!

As a first time submitter, I've given this a pretty tough first pass review, but it will need re-reviewing once you've addressed the points within, since the diff should be smaller then.

src/bech32.c
fadedreamz: This code path should never hit. So, having an assert actually helps with the debugging if somehow it occurs in the execution. But for now, to match with wally's error handling, I am just returning a placeholder value.

fadedreamz:
@jgriffiths Thank you for your review, much appreciated. I have more or less addressed all of your concerns. Please kindly review the patch again.

src/bech32.c
```diff
@@ -37,20 +42,26 @@ static uint32_t bech32_polymod_step(uint32_t pre) {
...
     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
     15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
     -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
-    1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
     -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
-    1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
```
jgriffiths, line 61: Please remove these whitespace changes

src/bech32.c
```diff
@@ -184,10 +202,13 @@ static int segwit_addr_decode(int *witver, uint8_t *witdata, size_t *witdata_len
     uint8_t data[84];
     char hrp_actual[84];
     size_t data_len;
-    if (!bech32_decode(hrp_actual, data, &data_len, addr, 90)) goto fail;
+    bech32_encoding enc = BECH32_ENCODING_NONE;
```
jgriffiths, line 205: declare this with the other vars at the start of the function, and remove the unneeded initialisation.

src/bech32.c
```diff
@@ -206,30 +227,34 @@ int wally_addr_segwit_from_bytes(const unsigned char *bytes, size_t bytes_len,
...
         *output = 0;
 
     if (!addr_family || flags || !bytes || !bytes_len || !output)
         return WALLY_EINVAL;
 
-    if (bytes[0] != 0)
-        return WALLY_EINVAL; /* Only v0 witness programs are currently allowed */
+    witver = bytes[0];
+    if (witver != 0)
+        witver -= 0x50;
```
jgriffiths, line 240: This needs some explanation, why are we subtracting 0x50 here?

src/bech32.c
```diff
@@ -206,30 +227,34 @@ int wally_addr_segwit_from_bytes(const unsigned char *bytes, size_t bytes_len,
...
-    if (!segwit_addr_encode(result, addr_family, 0, bytes + 2, bytes_len - 2))
+    if (!segwit_addr_encode(result, addr_family, witver, bytes + 2, bytes_len - 2))
         return WALLY_ERROR;
 
     *output = wally_strdup(result);
     wally_clear(result, sizeof(result));
     return *output ? WALLY_OK : WALLY_ENOMEM;
 }
 
-
```
jgriffiths, line 232: Please remove this whitespace change

src/bech32.c
```diff
@@ -37,20 +42,26 @@ static uint32_t bech32_polymod_step(uint32_t pre) {
            (-((b >> 4) & 1) & 0x2a1462b3UL);
 }
 
-static const char *charset = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+static uint32_t bech32_final_constant(bech32_encoding enc) {
+    if (enc == BECH32_ENCODING_BECH32) return 1;
+    if (enc == BECH32_ENCODING_BECH32M) return 0x2bc830a3;
+    return 0xbad; /* Should never reach this path */
+}
```
jgriffiths, line 49: This function can be removed by using the constants directly as above

src/bech32.c
```diff
@@ -78,36 +89,36 @@ static int bech32_encode(char *output, const char *hrp, const uint8_t *data, siz
     for (i = 0; i < 6; ++i) {
         chk = bech32_polymod_step(chk);
     }
-    chk ^= 1;
+    chk ^= bech32_final_constant(enc);
```
jgriffiths, line 92: pass the checksum constant directly to this function and use it here, there is then no need to reason about the failure case in bech32_final_constant().

src/bech32.c
```diff
@@ -138,9 +149,15 @@ static int bech32_decode(char *hrp, uint8_t *data, size_t *data_len, const char
...
     if (have_lower && have_upper) {
-        return 0;
+        return BECH32_ENCODING_NONE;
+    }
+    if (chk == bech32_final_constant(BECH32_ENCODING_BECH32)) {
+        return BECH32_ENCODING_BECH32;
+    } else if (chk == bech32_final_constant(BECH32_ENCODING_BECH32M)) {
+        return BECH32_ENCODING_BECH32M;
+    } else {
+        return BECH32_ENCODING_NONE;
```
jgriffiths, line 159: return the checksum constant directly, or e.g. BECH32_CHECKSUM_UNKNOWN if invalid

src/bech32.c
```diff
@@ -26,6 +26,11 @@
 #include <include/wally_crypto.h>
 #include "script.h"
 
+typedef enum {
+    BECH32_ENCODING_NONE,
+    BECH32_ENCODING_BECH32,
+    BECH32_ENCODING_BECH32M
+} bech32_encoding;
```
jgriffiths, line 33: Instead of an enum for the encoding, use a single enum for the checksum and use that to differentiate the encoding (with e.g. BECH32_CHECKSUM_UNKNOWN being 0 representing BECH32_ENCODING_NONE)

src/bech32.c
```diff
@@ -168,13 +185,14 @@ static int convert_bits(uint8_t *out, size_t *outlen, int outbits, const uint8_t
 static int segwit_addr_encode(char *output, const char *hrp, int witver, const uint8_t *witprog, size_t witprog_len) {
     uint8_t data[65];
     size_t datalen = 0;
+    bech32_encoding enc = witver ? BECH32_ENCODING_BECH32M : BECH32_ENCODING_BECH32;
```
jgriffiths, line 188: use the checksum constant here and make the variable const

src/test/test_bech32.py
```diff
@@ -6,14 +6,18 @@
         ["bc", 0, '0014751e76e8199196d454941c45d1b3a323f1433bd6'],
     'tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sl5k7':
         ["tb", 0, '00201863143c14c5166804bd19203356da136c985678cd4d27a1b8c6329604903262'],
-    'bc1pw508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7k7grplx':
```
jgriffiths, line 9: Please add this to failure cases with the comment that v1 must be bech32m encoded

jgriffiths (commented):
Hi @fadedreamz Please take a look at the comments, let me know if you need something clarified, thanks!

src/bech32.c
fadedreamz: OP_0 is 0x0, and OP_1 .. OP_16 is 0x51 to 0x60. So, if it is not OP_0, then we can get the version by subtracting 0x50 from the value.

Sjors: If you go that route, adding a comment would help. But you'd also have to check that `witver` > `0x50` to prevent underflow.

It'll be years before there's a v2 witness version, so just checking explictly for `OP_0` and `OP_1` might be easier.

Sjors (commented):
80afdd2 looks mostly correct to me

src/bech32.c
jgriffiths: This is what I meant above, we should just be checking for the allowed versions OP_0 and OP_1 (disregarding the core advice to allow unknown versions since as v0 to v1 showed us, there is no guarantee this will work without further changes).

fadedreamz: @Sjors The next line there is a check that will prevent any underflow to pass.
```
if (witver > 1) /* For now we only support v0 and v1 addresses */
        return WALLY_EINVAL;
```
In fact, it will only allow the exact value of witver that is in {0,1}. As the passed byte is `uchar` and the witver is taken as `uint8_t` this essentially guarantees that only OP_0 and OP_1 will be passed. I wanted to keep this code instead of exactly checking OP_0 and OP_1 because in the future if OP_2 needs to be supported and it uses bech32m then all we need to do will be to change the comparison line.
But if you and @jgriffiths think it is better to check OP_0 and OP_1 explicitly I will change it to that. Let me know.

jgriffiths: I prefer the explicit comparison because it avoids magic numbers in the code. Also its likely that when the next version is added we may have further logic to check after verifying the version, so keeping the version check a single 'if' in the future may well not be possible anyway.

fadedreamz: @jgriffiths I have added explicit comparison. Kindly review the patch.

fadedreamz:
@Sjors @jgriffiths I have added explicit checks. Kindly review it again. Thanks.

src/bech32.c
jgriffiths: Thanks @fadedreamz - I am a little caught up with some gdk changes so it might be a while before I can look at this, please know it's on my list to action!

src/bech32.c
```diff
@@ -206,22 +218,26 @@ int wally_addr_segwit_from_bytes(const unsigned char *bytes, size_t bytes_len,
...
     if (output)
         *output = 0;
 
     if (!addr_family || flags || !bytes || !bytes_len || !output)
         return WALLY_EINVAL;
 
-    if (bytes[0] != 0)
-        return WALLY_EINVAL; /* Only v0 witness programs are currently allowed */
+    witver = bytes[0];
+    if (witver != OP_0 && witver != OP_1)
```
Sjors, line 230: You lost the comment here:
```c
if (witver != OP_0 && witver != OP_1) {
  // Only v0 and v1 witness programs are currently allowed
```

src/bech32.c
```diff
@@ -206,22 +218,26 @@ int wally_addr_segwit_from_bytes(const unsigned char *bytes, size_t bytes_len,
...
+        return WALLY_EINVAL;
+        
+    witver &= 0x0F; /* Convert OP_0 -> 0, OP_1 -> 1 etc */
 
     ret = script_get_push_size_from_bytes(bytes + 1, bytes_len - 1, &push_size);
-    if (ret != WALLY_OK || (push_size != HASH160_LEN && push_size != SHA256_LEN))
+    if (ret != WALLY_OK || (witver == 0 && push_size != HASH160_LEN && push_size != SHA256_LEN))
```
Sjors, line 236: For Taproot you can add this constraint (maybe move `ret != WALLY_OK` to a separate check):
```c
if (ret != WALLY_OK || (witver == 1 && push_size != SHA256_LEN))
```

Sjors (commented):
ACK 018062f modulo missing v1 length check and conditional on CI

jgriffiths:
@benma yes, its on my TODO list along with the outstanding changes as soon as my current  gdk work is committed.

jgriffiths:
@fadedreamz My apologies for the delay in getting to this. In the end I have taken some of your work from this PR into https://github.com/ElementsProject/libwally-core/pull/297 which implements bech32/blech32 plus a bunch of other stuff I have been meaning to merge for a long time.

I have credited you on the bech32 commit in the PR, if you'd like to review it you are more than welcome (Note I ended up relaxing the version requirements after re-reading the BIPs).

Thanks for your contribution and once again sorry for the delay in addressing this PR.

PR #291, opened 2021-09-04
removed a trailing tick

sigkill:
There was a trailing character in the readme "`" that was causing some issue when copying some of the text.


jgriffiths:
@sigkill I will merge this shortly, thanks!

PR #297, opened 2021-10-15
Crypto Garage merge

jgriffiths:
This is the first part of merging diverged changes from https://github.com/cryptogarageinc/ (initially, the cfd-develop branch), with fixes, cleanups and test updates applied.

@k-matsuzawa If you could review I would be grateful, I will be adding your output descriptor changes next.


k-matsuzawa:
ACK ff428db  (I confirmed that all tests succeed locally.)

src/script.c
```diff
@@ -933,27 +944,27 @@ int wally_varbuff_to_bytes(const unsigned char *bytes, size_t bytes_len,
...
         return WALLY_EINVAL;
 
     if (flags & ALL_SCRIPT_HASH_FLAGS) {
         if (!bytes_len)
             return WALLY_EINVAL;
-    } else if (bytes_len != HASH160_LEN && bytes_len != SHA256_LEN) {
-        /* Only v0 witness scripts are currently supported */
-        return WALLY_EINVAL;
+    } else if (version == 0 && bytes_len != HASH160_LEN && bytes_len != SHA256_LEN) {
+        return WALLY_EINVAL; /* Invalid length for v0 witness script */
```
LeoComandini, line 967: Should we check that `bytes_len` is less than or equal to 40 if version is not 0?

src/test/test_confidential_addr.py
```diff
@@ -7,6 +7,47 @@
 CA_PREFIX_LIQUID_REGTEST = 0x04
 EC_PUBLIC_KEY_LEN = 33
 
+hrp = lambda s: s[:s.find('1')]
+
+#   pk=02409b4d18429c6e5cbc0bd59c63b8fe7055f603190c8deed6a644bc95c9772e48
+#   script=OP_HASH160 dc4af3ea14b0592621514e2bd4a0e083c7fac2f2 OP_EQUALVERIFY
+#   -> a914dc4af3ea14b0592621514e2bd4a0e083c7fac2f288
+#   tr hash=409b4d18429c6e5cbc0bd59c63b8fe7055f603190c8deed6a644bc95c9772e48
+segwit_valid_cases = [
```
LeoComandini, line 16: We have some test vectors [here](https://github.com/ElementsProject/elements/blob/master/test/functional/rpc_invalid_address_message.py) if we want to add test coverage also for invalid cases, but it can be done later if you prefer

jgriffiths: Noted, thanks. I think I will add further liquid tests in another commit unrelated to this one, in parallel with enabling taproot sends on liquid in gdk (this PR has been tested with BTC/gdk to ensure taproot sends work as expected).

jgriffiths:
> For sanity preserving reasons, maybe point out that this is not actually a valid taproot address.

@Sjors I'm not sure which line of code you are referring to here?

PR #301, opened 2021-10-21
Add --enable-standard-secp configure argument to build with a non-zkp…

jgriffiths:
… libsecp


configure.ac
```diff
@@ -168,6 +172,12 @@ fi
 if test "x$elements" == "xyes"; then
     AX_CHECK_COMPILE_FLAG([-DBUILD_ELEMENTS=1], [AM_CFLAGS="$AM_CFLAGS -DBUILD_ELEMENTS=1"])
 fi
+if test "x$standard_secp" == "xyes"; then
+    if test "x$elements" == "xyes"; then
+        AC_MSG_FAILURE([ERROR: Elements cannot be enabled with standard libsecp256k1])
+    fi
+    AX_CHECK_COMPILE_FLAG([-DBUILD_STANDARD_SECP=1], [AM_CFLAGS="$AM_CFLAGS -DSTANDARD_SECP=1"])
```
Sjors, line 179: Missing `BUILD_` at the end.

Sjors (commented):
Maybe also add the choice to the build options summary.

The configure check against combining `--enable-standard-secp --enable-elements` works.

However internal.h still tries to include `#include "secp256k1/include/secp256k1_ecdsa_s2c.h"` despite the `#ifndef BUILD_STANDARD_SECP`. It always seems to be defined. See inline for fix.

There's also a bunch of includes in `elements.c` (above `#ifdef BUILD_ELEMENTS`) that need to be excluded. As well as two functions in `sign.c`. See https://github.com/Sjors/libwally-core/commits/updates

After that it builds and the (limited) test suite in LibWallySwift still passes.

Sjors:
tACK 7e010f327a707c81b55cc2756fb894a3e763d9c0

PR #309, opened 2021-12-15
bip32: Add support for key derivation from BIP32 style path strings

jgriffiths:
Required for descriptor support.


src/bip32.c
```diff
@@ -80,6 +85,90 @@ static bool version_is_mainnet(uint32_t ver)
...
+        goto fail;
+
+    *written = 0;
+
+    if (flags & BIP32_FLAG_STR_BARE) {
+        if (i < str_len && str[i] == '/')
+            goto fail; /* bare path must start with a number */
+    } else {
+        if (i < str_len && (str[i] == 'm' || str[i] == 'M'))
+            ++i; /* Skip */
```
LeoComandini, line 111: we should either document that `m` or `M` are ignored and `flags` is used to derive privately or not, or we should expect that those are consistent, or that `m` or `M` if present overrides `BIP32_FLAG_KEY_P*`

LeoComandini:
utACK e01c3b7ccbff6a4e8d8cbb18bf15ea979d4dbfa3, with comment

src/bip32.c
jgriffiths: I'm unable to find any references to `m`/`M` being used to control the final derivation type, rather than it and the hardened indicator `h`/`H` being just case-insensitive. Do you have a link to any docs/software that behaves this way?

The upcoming descriptors PR includes a bip32 change to allow public derivation where hardened path elements are present in the path and we are starting with a non-neutered key. If you can point to any general use of the master key case to control the derivation type, I'm happy to allow inferring it from the input in that PR (perhaps controlled by a flag).

LeoComandini: From https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#the-key-tree,

    To shorten notation, we will write CKDpriv(CKDpriv(CKDpriv(m,3H),2),5) as m/3H/2/5.
    Equivalently for public keys, we write CKDpub(CKDpub(CKDpub(M,3),2),5) as M/3/2/5
    ...
    N(m/a/b/c) = N(m/a/b)/c = N(m/a)/b/c = N(m)/a/b/c = M/a/b/c

As a caller, if I pass `xprv...` and `M/1` I'd expect a neutered key in the output.
I think we should document it to make explicit the behavior we want.

Another option is to not allow the leading `m`/`M` which make things a bit ambiguous, since those functions are used for descriptors (so in the form `xpub/0/*`).

jgriffiths: OK, I think I see the misunderstanding.

The spec only defines one result of derivation, which depends on the type of key you start with. Per the table in bip32.c, it doesn't describe deriving e.g. a pub key from a priv key, instead it assumes that if you start with a priv key, you will derive the priv key and then convert the result to a pub key (neutering it). The upper case notation here is just a shorthand to express the derivation identities.

Note that the `m`/'M' in this notation just defines the starting key type, it does not define the desired eventual result; that is defined by the starting key type _plus_ whether there is a surrounding `N()'. Since there is no `N`-notation in the bip32 path strings, the path strings cannot express the desired derived key type using the case of 'm'. This is why no existing libraries interpret the leading 'm' to mean this. Instead you either see priv->pub derivation forced to go via converting from the priv key (e.g rust bip32), or some other method like https://github.com/richardkiss/pycoin/blob/afe30cd45802c44a160cc74779a6ce047b2ece70/pycoin/key/BIP32Node.py#L198 which uses a `.pub` suffix, or in wallys case, we use the derivation flags.

Various libraries allow upper, lower or mixed case strings with little consistency _except_ that none of them assign meaning to the case of the `m`. I think I will add a flag to ignore case, otherwise force lower case. Testing with descriptors against core it seems that upper case hardening `H` indicators are not allowed, and rust bip32 only allows lower case `m` and doesn't allow `h`/'H' at all. A flags seems to be the best way to control that behaviour given we want to support as many inputs as possible, but also need to validate identically to core for descriptors.

PR #202, opened 2020-06-02
feat: output descriptor

k-matsuzawa:
Since I wanted to parse miniscript, I also implemented a function to parse the output descriptor (see #151 ).
I have confirmed the execution by ctypes in Python, but not in Java. Please check the API parameters are correct for using swig. :pray:

PR #310, opened 2021-12-17
Output descriptors/miniscript support

jgriffiths:
Based on the excellent work of  @k-matsuzawa in https://github.com/ElementsProject/libwally-core/pull/202 (which this PR replaces).

The major changes from the original PR are:
- Move to a parse-first API to support multiple operations/queries without re-parsing.
- Use `wally_map` for variable substitutions.
- Use the new `_n` api call variants to work directly from the input string
- Remove all sub-allocations while parsing
- Remove script generation allocations completely in most cases, otherwise from 1MB to typically less than 1k.
- Expose descriptor canonicalization, network and features
- API support for future ranged descriptors
- Expose to SWIG targets
- Extend the test cases
- Various bug fixes, speedups, simplifications

@k-matsuzawa I would be very grateful if you could review when you have time.

Note: This is rebased on https://github.com/ElementsProject/libwally-core/pull/365, and will be updated until that branch is merged to master. Only the last commit is specific to this PR.


jgriffiths:
Note that the parsing code still allocates a large buffer for each child when generating scripts; I intend to fix this by computing the exact required length up-front in the future. Since this is an internal change, it won't affect the exposed API/ABI.

darosior:
I still plan to review this. (Sorry for taking so long with #202..)

src/descriptor.c
```diff
@@ -0,0 +1,2323 @@
...
+    while (*p && *p != '#') {
+        while (*p && *p != '#' && !is_identifer_char(*p)) {
+            ++required_len;
+            ++p;
+        }
+        start = p;
+        while (is_identifer_char(*p))
+            ++p;
+        if (p != start) {
+            const bool is_number = *start >= '0' && *start <= '9';
```
k-matsuzawa, line 2275: It seems that only the beginning of the buffer is being checked. I think you should check the whole range of lookup_len.

k-matsuzawa (commented):
Thanks! I reviewed it.
I commented on one thing. Other than that it looks good!

>Note that the parsing code still allocates a large buffer for each child when generating scripts; I intend to fix this by computing the exact required length up-front in the future. Since this is an internal change, it won't affect the exposed API/ABI.

I'm sorry, please take care of it. 🙏

src/descriptor.c
jgriffiths: @k-matsuzawa This check just looks at the first digit because variable identifiers must start with a letter or underscore; if the string of characters starts with a number (or if its too long to be an identifier) we avoid looking it up for variable substitution. So for example, `key_1` is a valid variable name, but `1_key` is not (variables use the same naming constraints as e.g. python, c and c++, as well as additionally being limited to 16 characters in length).

I'll rename the variable to `starts_with_digit` and update the comment to make this clearer, thanks!

jgriffiths:
> I'm sorry.

The implementation is totally suitable for typical use, please don't apologise! My desire to reduce these allocations is only for the special case of using wally on constrained devices like Jade and the bitbox02, but I don't expect external contributors to have to optimise for that when submitting!

Sjors:
Would it make sense to split out Miniscript parsing and do separately? I though there were still some pending changes to the Miniscript syntax to make it compatible with descriptors? https://github.com/bitcoin/bitcoin/pull/16800#issuecomment-903211605

darosior:
The Miniscript syntax for P2WSH is now final. The `pkh()` and `pk()` aliases, as well as the rename of `thresh_m()` to `multi()` made the syntax of Miniscript descriptors backward compatible with "legacy" descriptors.

jgriffiths:
Note: Moving back to WIP, I think we need a couple more features.

src/bip32.c
```diff
@@ -629,14 +627,34 @@ int bip32_key_from_parent_path(const struct ext_key *hdkey,
...
     for (i = 0; i < child_path_len; ++i) {
         struct ext_key *derived = &tmp[tmp_idx];
+        uint32_t derivation_flags = flags;
+
+        if (private_until && i < private_until - 1) {
+            /* Derive privately until we reach the last hardened child */
+            derivation_flags &= ~BIP32_FLAG_KEY_PUBLIC;
+            derivation_flags |= BIP32_FLAG_KEY_PRIVATE;
+        }
+        if (i + 2 < child_path_len)
```
LeoComandini, line 651: Shouldn't this be `+ 1`?

LeoComandini:
utACK 99b25e063ed782aa737ee02674fe4d88eedae7d0 (just the first commit "bip32: Allow public path derivation with hardened child elements"), with comment

Sjors:
There's a couple of functions that I would find quite useful, for sanity checking a multisig wallet: 

* infer the (sub)descriptor type(s): e.g. it's a witness script hash (`KIND_DESCRIPTOR_SH`) with a sorted multisig (`KIND_DESCRIPTOR_MULTI_S `)
* for multisig: get the threshold value
* extract xpub(s)

All of that currently happens internally in `descriptor.c`. Could wait for a followup.

Sjors (commented):
I was able to wrap two of the new functions and run some rudimentary sanity checks on them in https://github.com/Sjors/libwally-swift/pull/57

include/wally_descriptor.h
```diff
@@ -0,0 +1,145 @@
...
+ * :param child_num: The BIP32 child number to derive.
+ * :param network: ``WALLY_NETWORK_`` constant descripting the network to generate for.
+ * :param flags: For future use. Must be 0.
+ * :param output: Destination for the resulting addresss.
+ *|    The string returned should be freed using `wally_free_string`.
+ */
+WALLY_CORE_API int wally_descriptor_to_address(
+    const char *descriptor,
+    const struct wally_map *vars_in,
+    uint32_t child_num,
```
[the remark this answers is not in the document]
jgriffiths: This seems like a reasonable restriction, agreed.

include/wally_descriptor.h
```diff
@@ -0,0 +1,145 @@
...
+ *|    The string returned should be freed using `wally_free_string`.
+ *
+ * Replaces any variables from ``vars_in`` with their mapped values,
+ * and adds a checksum if required. Key names for ``vars_in`` must be 16
+ * characters or less and start with a letter.
+ *
+ * .. note:: Other canonicalization (hardened derivation indicator
+ * mapping, and private to public key mapping) is not yet implemented.
+ */
+WALLY_CORE_API int wally_descriptor_canonicalize(
```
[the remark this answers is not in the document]
jgriffiths: Many of these checks would require full parsing of the descriptor, which is the responsibilitiy of the other functions. I originally intended this call to allow full canonicalisation of the input (for example, sortedmulti with x,y,z is the same as x,z,y or other permutations), however it turns out that core descriptors have no such concept, and do not compute checksums on a canonical representation which puts limits on the kinds of cleanup that are useful in practice, and the ability to compare descriptors when they can be trivially permuted.

However the primary reason I moved this back to WIP status is the script generation, which will also need interface changes, so I will re-review this function as part of that.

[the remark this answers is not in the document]
Sjors: You could also add a flag to perform (or skip) full validation, or have another method for it. But in practice both validating and getting the canonical version are something any safe software should probably do on data from the outside. Otherwise software has to delay error handling until some later step.

While you're changing things: can you add a flag to mandate a checksum, for all functions? I found myself parsing the string for `#`. The current code already ensures that _if_ a checksum is present it's correct, but it doesn't care if a checksum is missing.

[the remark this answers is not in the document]
jgriffiths: >  I found myself parsing the string for #.

You can canonicalize the string and then compare it to your input, if canonicalization fails the input is invalid, if it succeeds and the result differs then the original had no checksum.

I'll consider adding a flag to validate.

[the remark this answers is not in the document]
Sjors: Doesn't canonicalization change `h` to ` and thus change the checksum?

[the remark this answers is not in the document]
jgriffiths: >Doesn't canonicalization change h to ` and thus potentially change the checksum?

No, all canonicalization currently does is variable substitution and append a checksum if missing.

Sjors:
Similar to what I'm suggesting in https://github.com/bitcoin/bitcoin/pull/24147#issuecomment-1029278730, instead of introducing all of Miniscript you could start with the subset that's needed for current descriptor functionality: `KIND_MINISCRIPT_PK(_K)`, `KIND_MINISCRIPT_PK(H/_H)`, and `KIND_MINISCRIPT_MULTI`. You only need `TYPE_B` and `TYPE_K` and none of the `PROP_` logic.

jgriffiths:
> For the Miniscript part, how about we look into making https://github.com/sipa/miniscript/ a library and exposing a C API?

C++ code is not acceptable for wally, and our implementation requirements are stricter than some other implementations; because we need to work on embedded devices and hardware wallets we need to constrain both code size and stack/memory usage for exposed functionality.

darosior:
Is it still WIP? What is left to be done there? Any way i can help this work to move forward?

jgriffiths:
> Is it still WIP? What is left to be done there? Any way i can help this work to move forward?

Hi @darosior sorry for the late reply. This PR is still WIP for 2 reasons:

1 - The interface currently only allows generating one script type; I'd like to extend this to be able to generate scriptsigs, scriptpubkeys and witness scripts (as well as tapscript in due course)
2 - The creation of scripts internally currently requires a lot of memory which needs fixing in order to work well on smaller devices like the Jade or Bitbox02

Item 1 is a pre-requisite for merging as it will change the API interface slightly. 2 is not  blocker for merging as the internals can be refactored once the interface is stable.

I am currently working on PSBT v2/PSET support for wally, and hope to be able to re-visit this following that. Unfortunately I have rather extreme demands on my time at present.

src/bip32.c
jgriffiths: Note the bip32 changes are being merged separately, but no, the `+ 2` is correct. This is equivalent to `i + 1 < child_path_len - 1` (zero based index vs array size). You can verify the behaviour is correct by changing this and observing the resulting test failures.

Loopsbetas2173 (approved):

Loopsbetas2173 (commented):
- - @jgriffiths

include/wally_descriptor.h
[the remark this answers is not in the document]
jgriffiths: done

include/wally_descriptor.h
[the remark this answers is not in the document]
jgriffiths: updated to a parse-then-query API so resolving this

jgriffiths:
Hi @k-matsuzawa my apologies for taking so long to get back to this PR. I have updated it further, including changing the API, and hope to get this into the upcoming 0.8.8 release.

I hope you have have time to review this new version, and will switch to it in due course. In the event that you will remain with your own implementation for a time, please see the individual commits squashed for this PR at https://github.com/ElementsProject/libwally-core/tree/descriptor_backup_7_feb_23. There are some bug fixes there which you may wish to pick up.

Compared to the initial version, this now uses a parsed representation as the descriptor object rather than passing around the string. This is needed to prevent re-parsing overhead for multiple operations/queries on the descriptor. The large upfront memory allocation for script generation has been removed. In most cases no allocations are now needed to generate. When allocation is required, the length allocated is reduced to the maximum length needed to contain all sub-scripts, which is typically in the order of 1kb or so.

I have also merged in your branches extra fixes as you can see from the link above.

k-matsuzawa:
@jgriffiths Thanks!
I was able to quickly check out the latest implementation and confirm that it works the same way. I will replace it when it is released 0.8.8!

jgriffiths:
@k-matsuzawa thanks, looking forward to merging this shortly :)

jgriffiths:
@Sjors FYI

Sjors:
Nice! It looks like the first 5 commits could be split off, so review can focus on the actual descriptors / miniscript stuff?

include/wally_descriptor.h
```diff
@@ -0,0 +1,172 @@
...
+ * .. note:: This function overestimates the script size required, but is
+ *|    cheap to compute (does not require script generation).
+ */
+WALLY_CORE_API int wally_descriptor_to_script_get_maximum_length(
+    const struct wally_descriptor *descriptor,
+    uint32_t flags,
+    size_t *written);
+
+/**
+ * Create a script corresponding to an output descriptor or miniscript expression.
```
LeoComandini, line 108: Which script is created? 

rust-miniscript allows to generate scriptpubkey, (unsigned) script sig, explicit script/witness script/redeem script, script code, etc. I guess here it's the scriptpubkey, if so I think it's worth mentioning it in the docs and perhaps in the function(s) name

include/wally_descriptor.h
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * :param descriptor: Parsed output descriptor or miniscript expression.
+ * :param depth: Depth of the expression tree to generate from. Pass 0 to generate from the root.
+ * :param index: The zero-based index of the child at depth ``depth`` to generate from.
+ * :param variant: The variant of descriptor to generate. Pass 0 for the default.
```
LeoComandini, line 113: What is `variant` reserved for?

include/wally_descriptor.h
LeoComandini: So IIUC you can control the ("sub")script generated with `depth` and `index`. It wasn't very clear to me without looking at the tests, perhaps it's worth expanding a bit the documentation here.

include/wally_descriptor.h
```diff
@@ -0,0 +1,172 @@
...
+ * :param flags: Include ``WALLY_MINISCRIPT_ONLY`` to disallow descriptor
+ *|    expressions, ``WALLY_MINISCRIPT_TAPSCRIPT`` to use x-only pubkeys, or 0.
+ * :param output: Destination for the resulting parsed descriptor.
+ *|    The descriptor returned should be freed using `wally_descriptor_free`.
+ *
+ * Variable names can be used in the descriptor string and will be substituted
+ * with the contents of ``vars_in`` during parsing. Key names for ``vars_in``
+ * must be 16 characters or less in length and start with a letter.
+ */
+WALLY_CORE_API int wally_descriptor_parse(
```
Sjors, line 38: This could use improved error messages / codes. For example I'm getting a `-2` (`WALLY_EINVAL`) when parsing `pkh(xpub68Gmy5EdvgibQVfPdqkBBCHxA5htiqg55crXYuXoQRKfDBFA1WEjWgP6LHhwBZeNK1VTsfTFUHCdrfp1bgwQ9xv5ski8PX9rL2dZXvgGDnw/1)` with network set to `WALLY_NETWORK_BITCOIN_MAINNET`, `flags` set to zero and variables to `NULL`.

jgriffiths:
> Nice! It looks like the first 5 commits could be split off, so review can focus on the actual descriptors / miniscript stuff?

Yes, sorry if it isn't clear, the first commits are a separate and will be merged to master shortly, I'm keeping this PR rebased over master for easier testing by reviewers.

Sjors:
Can you update the PR description to point to #365? That seems to be where those commits are living.

include/wally_descriptor.h
jgriffiths: `variant` has been added to support descriptors that return multiple values, the only example of which currently is `combo`. When generating a script/address/address range you normally just have a child number and a bijection between that child number and a given address. The descriptor specs make it clear that to support this bijection, all ranged keys (those with '/*') must be derived at the same child number in a descriptor expression.

 `combo`, along with sub-range key-expressions like rust-bitcoins `<0;1>` extension, break this model and the bijection, which is unfortunate. I consider this a misfeature given (a) there is no specification of the order in which these addresses should be iterated/returned and (b) there is otherwise no mechanism to return them for languages which do no support variant(unitary/list) return types.

Nonetheless we must support it, and `variant` is the mechanism that will allow this. The idea is that each expression construct that breaks the bijection is a variant, and the parsed expression can be queried for the number of variants it contains. variant + child number then biject to a single script/address, with the order documented per-expression type (for `combo` this would be `0=P2PK`, `1=P2PKH`, `2=P2WPKH`, or `3=P2SH-P2WPKH` for example).

include/wally_descriptor.h
```diff
@@ -0,0 +1,172 @@
...
+    uint32_t flags,
+    unsigned char *bytes_out,
+    size_t len,
+    size_t *written);
+
+/**
+ * Create an address corresponding to an output descriptor.
+ *
+ * :param descriptor: Parsed output descriptor.
+ * :param variant: The variant of descriptor to generate. Pass 0 for the default.
```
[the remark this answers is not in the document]
jgriffiths: Please see my comment above. Note that variants are WIP, I had planned to add these to the release notes but will also add source comments.

jgriffiths:
> Can you update the PR description to point to #365?
Done, sorry for the confusion.

include/wally_descriptor.h
jgriffiths: I've just added this as a test case and it parses without error, please see the commit.

For error handling, I agree both descriptors and PSBT support demand better error messages. The intention here is to support (optional) rich errors for all calls with a `wally_get_error_message` API, but this will be a separate undertaking/PR.

include/wally_descriptor.h
jgriffiths: Added documentation notes for both. Generating other script types in the future will be handled via the `flags` argument rather than separate functions for each type.

include/wally_descriptor.h
Sjors: > we must support it

Do we? `combo` descriptors are only supported by Bitcoin Core for its own legacy wallet backwards compatibility. Each descriptor type has it's own BIP number, so it's possible to simply not support [BIP 384](https://github.com/bitcoin/bips/blob/master/bip-0384.mediawiki). I'm not familiar with rust-bitcoins's `<0;1>` extension.

include/wally_descriptor.h
Sjors: I'm not seeing it in `src/test/test_descriptor.py` in 4d430b8182d58387453889dec291450a4ab65cd5. Did you mean to push a new version?

But if it passes, maybe I'm doing something wrong on the Swift - C mapping...

include/wally_descriptor.h
jgriffiths: I think we have to support variants, yes. combo is useful for watching/sweeping legacy addresses, and the rust extension suggests there may be a bip44 compatible operator proposed in the future that supports the 0,1 change address delineation when generating addresses (or if not, that ranges themselves will be pushed forward to support this use case).

include/wally_descriptor.h
jgriffiths: Ah yes I haven't pushed this yet, sorry. I will shortly.

include/wally_descriptor.h
darosior: Just a note that this is not that BIP389 descriptors aren't a rust-bitcoin extenstion: https://github.com/bitcoin/bips/pull/1354. I share the sentiment wrt it, but it was a compromise for backups that i now find reasonable.

jgriffiths: I've pushed full support for `combo()` using variants to demonstrate how this works. I will merge these commits back into the main one for merging.

jgriffiths: @k-matsuzawa Please note that this changes the behaviour of `combo()` which was previously not quite correct (in that there was no way to generate all variants).

You can now query the number of variants an expression has, and generate scripts/addresses for each variant. Note that for `combo()`, the p2pk variant (0) generates scripts but errors when generating addresses, as p2pk has no address format.

include/wally_descriptor.h
jgriffiths: pushed, as commented above

include/wally_descriptor.h
jgriffiths: > Just a note that BIP389 descriptors (multipath key expressions) aren't a rust-bitcoin extenstion: [bitcoin/bips#1354](https://github.com/bitcoin/bips/pull/1354). I share the sentiment wrt it, but it was a compromise for backups that i now find reasonable.

Thanks, I wasn't aware of this BIP, I've added review comments to it. I think its probably worth adding the range item to the exposed API for when they are eventually implemented in wally.

include/wally_descriptor.h
Sjors: Thanks, it was bug on my end.

Sjors (commented):
I was able to rebase my LibWallySwift implementation on the latest changes. It uses only a small fraction of the functionality though.

Some features that are still on my wish list, but can wait for a followup:
1. Determine the descriptor type (e.g. `sorted_multi()`) and output type (legacy, p2sh segwit, native segwit, taproot)
2. Obtain the (extended) keys
3. Obtain the threshold for (sorted_multi()) and other such variables

include/wally_descriptor.h
jgriffiths: > its probably worth adding the range item to the exposed API

Done.

jgriffiths:
> infer the network (though you can't distinguish testnet, regtest and signet, and you can't tell from a hex public key, so maybe this is not realistic)

@Sjors its turns out this is required in order to allow address generation from untyped expressions. Please see the latest commits around `get_network`/`set_network` for details.

include/wally_descriptor.h
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Parse an output descriptor or miniscript expression.
+ *
+ * :param descriptor: Output descriptor or miniscript expression to parse.
+ * :param vars_in: Map of variable names to values, or NULL.
+ * :param network: ``WALLY_NETWORK_`` constant descripting the network the
+ *|    descriptor belongs to, or WALLY_NETWORK_NONE for miniscript-only expressions.
+ * :param flags: Include ``WALLY_MINISCRIPT_ONLY`` to disallow descriptor
+ *|    expressions, ``WALLY_MINISCRIPT_TAPSCRIPT`` to use x-only pubkeys, or 0.
```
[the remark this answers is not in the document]
jgriffiths: Sure, added.

jgriffiths:
OK, I've squashed the recent commits in this branch and backed up the un-squashed one in the [descriptor_backup](https://github.com/ElementsProject/libwally-core/tree/descriptor_backup) tree for review/posterity. This includes a new change to make use of a parsed descriptor thread-safe.

That's all the time I have for this release. AFAICS the API is now sufficient to support future work like multi-path and @Sjors wish list items.

@Sjors Please raise a bug to track your wish list so they aren't lost when this is merged, thanks.

PR #329, opened 2022-04-28
Build fixes

jgriffiths:
Updates the android build helpers to the latest supported ndk releases r22+ (LTS release r23b in particular).


JamieDriver:
Looks ok (although I'm no expert) - quick local test with ndk v23b and v24, both compiled without error.  
utack 3be40120

PR #330, opened 2022-06-02
PSBT v2

jgriffiths:


include/wally_elements.h
```diff
@@ -17,15 +17,20 @@ extern "C" {
...
- * Create a blinded Asset Generator from an Asset Tag and Asset Blinding Factor.
+ * Create an Asset Generator from an either an asset commitment or asset tag plus blinding factor.
  *
- * :param asset: Asset Tag to create a blinding generator for.
- * :param asset_len: Length of ``asset`` in bytes. Must be ``ASSET_TAG_LEN``.
- * :param abf: Asset Blinding Factor (Random entropy to blind with).
- * :param abf_len: Length of ``abf`` in bytes. Must be ``BLINDING_FACTOR_LEN``.
+ * :param asset: Asset Commitment or Tag to create a generator for.
+ * :param asset_len: Length of ``asset`` in bytes. Must be ``ASSET_COMMITMENT_LEN`` or ``ASSET_TAG_LEN``.
+ * :param abf: Asset Blinding Factor (Random entropy to blind with). Must be non-NULL when ``asset`` is a commitment.
```
JamieDriver, line 32: Must be NULL when asset is a commitment  (and non-NULL when it's a tag) ?

src/sign.c
```diff
@@ -315,6 +326,163 @@ int wally_ec_sig_to_public_key(const unsigned char *bytes, size_t bytes_len,
...
+    }
+
+    /* Check for addition of the scalars inverse */
+    memcpy(tmp, operand, len);
+    if (!seckey_negate(tmp))
+        return WALLY_ERROR; /* Outside the group order */
+
+    if (!memcmp(scalar, tmp, len)) {
+        /* X + -X = 0 */
+        wally_clear(bytes_out, len);
```
JamieDriver, line 382: Should be cleared already by op_args check, but no biggie (same in subtract below).  
Only mention because add and subtract explicitly clear, but multiply doesn't (and has a comment saying that it's already cleared).

src/elements.c
```diff
@@ -165,28 +216,37 @@ int wally_asset_rangeproof_with_nonce(uint64_t value,
...
-        get_generator(ctx, generator, generator_len, &gen) != WALLY_OK)
+        parse_generator(ctx, generator, generator_len, &gen) != WALLY_OK)
         goto cleanup;
 
     /* Create the rangeproof message */
-    memcpy(message, asset, ASSET_TAG_LEN);
-    memcpy(message + ASSET_TAG_LEN, abf, ASSET_TAG_LEN);
+    if (!is_explicit) {
+        memcpy(message, asset, ASSET_TAG_LEN);
+        memcpy(message + ASSET_TAG_LEN, abf, ASSET_TAG_LEN);
```
JamieDriver, line 242:
```suggestion
        memcpy(message + ASSET_TAG_LEN, abf, BLINDING_FACTOR_LEN);
```

JamieDriver:
Hey, I have had a read through, compiled it, played with the tests a bit - and raised a couple of small points.
It's nigh impossible to do a thorough review of this volume of change, but the overall impression is it's good and it does have a good amount of tests - so I'm in favour of putting it and 'fixing forward'.

jgriffiths:
Thanks @JamieDriver - appreciate the review effort, agreed it's a hard change to review. I'd like more tests tbh but I think they need to wait for in-progress Elements fixes before we can generate some better test cases.

This will live in master for a while before we release which should allow us to fix any issues found during external testing (especially via gdk).

jgriffiths:
@Sjors Would yo umind taking a look at these changes (at least, the interface changes for PSBT v0 in CHANGES.md)?

Please let me know if you find any issues or would like anything changed, thanks.

include/wally_crypto.h
```diff
@@ -482,6 +496,114 @@ WALLY_CORE_API int wally_ec_sig_to_public_key(
...
+ * Add one secp256k1 scalar to another.
+ *
+ * :param scalar: The starting scalar to have a value added to.
+ * :param scalar_len: The length of ``scalar`` in bytes. Must be ``EC_SCALAR_LEN``.
+ * :param operand: The scalar value to add to ``scalar``.
+ * :param operand_len: The length of ``operand`` in bytes. Must be ``EC_SCALAR_LEN``.
+ * :param bytes_out: Destination for the resulting scalar.
+ * :param len: The length of ``bytes_out`` in bytes. Must be ``EC_SCALAR_LEN``.
+ *
+ * .. note:: Computes (scalar + operand) % G. Returns ``WALLY_ERROR`` if
```
LeoComandini, line 519: `G` usually refers to the group generator, while `n` refers to the group order

include/wally_crypto.h
```diff
@@ -482,6 +496,114 @@ WALLY_CORE_API int wally_ec_sig_to_public_key(
...
+ *|    either input is not zero or within the secp256k1 group order G.
```
LeoComandini, line 520: Why do we fail if scalar is 0?

secp256k1 `scalar`s (sometimes referred to as `tweak`s) unlike secret keys can be 0. I think we should allow them to be 0.

include/wally_crypto.h
LeoComandini: Also below

include/wally_elements.h
```diff
@@ -17,15 +17,20 @@ extern "C" {
 
 #define ASSET_COMMITMENT_LEN 33 /** Length of an Asset Value Commitment */
 
-#define ASSET_RANGEPROOF_MAX_LEN 5134 /** Maximum length of an Asset Range Proof */
+#define ASSET_RANGEPROOF_MAX_LEN 5134 /** Maximum length of an Asset Value Range Proof */
+#define ASSET_EXPLICIT_RANGEPROOF_MAX_LEN 73 /** Maximum length of an Explicit Asset Value Range Proof */
```
LeoComandini, line 21: Is this the max length or the exact length?

in `src/elements.c` we have both
```
        if (asset || asset_len || abf || abf_len || extra || extra_len ||
            len < ASSET_EXPLICIT_RANGEPROOF_MAX_LEN)
            goto cleanup;
            
        ...
        
        if (!rangeproof || rangeproof_len > ASSET_EXPLICIT_RANGEPROOF_MAX_LEN ||
            parse_commitment(ctx, commitment, commitment_len, &commit) != WALLY_OK ||
            parse_generator(ctx, generator, generator_len, &gen) != WALLY_OK)
            return WALLY_EINVAL;
```

include/wally_elements.h
```diff
@@ -37,6 +42,28 @@ WALLY_CORE_API int wally_asset_generator_from_bytes(
     unsigned char *bytes_out,
     size_t len);
 
+/**
+ * Generate a rangeproof nonce hash using SHA256(ECDH(pub_key, priv_key).
+ *
+ * :param pub_key: Public blinding key.
```
LeoComandini, line 48: The shared secret used to encrypt data in the rangeproof is computed both ways
```
sha256(ecdh(receiver_public_blinding_key, sender_private_ephemeral_key)) == sha256(ecdh(sender_public_ephemeral_key, receiver_private_blinding_key))
```
I think the api documentation should reflect it, instead of describing only the sender way.

Also using `shared_nonce` instead of `nonce_hash` might be better (there is the `nonce` field but this is not simply its hash). But we might leave this name for consistency.

include/wally_elements.h
```diff
@@ -63,6 +90,26 @@ WALLY_CORE_API int wally_asset_final_vbf(
     unsigned char *bytes_out,
     size_t len);
 
+/**
+ * Compute the scalar offset used for blinding a confidential transaction.
+ *
+ * :param value: The transaction input value in satoshi.
```
LeoComandini, line 96:
```suggestion
 * :param value: Value in satoshi.
```

include/wally_elements.h
```diff
@@ -162,6 +210,46 @@ WALLY_CORE_API int wally_asset_rangeproof(
     size_t len,
     size_t *written);
 
+/**
+ * Generate an explicit value rangeproof for a transaction output.
+ *
+ * The nonce for this function should be randomly generated.
+ * See `wally_asset_rangeproof_with_nonce`.
+ */
+WALLY_CORE_API int wally_explicit_rangeproof(
```
LeoComandini, line 219: args documentation is missing

include/wally_elements.h
```diff
@@ -162,6 +210,46 @@ WALLY_CORE_API int wally_asset_rangeproof(
     size_t len,
     size_t *written);
 
+/**
+ * Generate an explicit value rangeproof for a transaction output.
```
LeoComandini, line 214: I'd remove the reference to "transaction output", an explicit rangeproof might be useful in more generic oontext.

I'd also add what this actually proves, i.e. that a value commits to a value commitment, without the need to reveal the value blinding factor to the verifier.

include/wally_elements.h
```diff
@@ -162,6 +210,46 @@ WALLY_CORE_API int wally_asset_rangeproof(
...
+ * The nonce for this function should be randomly generated.
+ * See `wally_asset_rangeproof_with_nonce`.
+ */
+WALLY_CORE_API int wally_explicit_rangeproof(
+    uint64_t value,
+    const unsigned char *nonce,
+    size_t nonce_len,
+    const unsigned char *vbf,
+    size_t vbf_len,
+    const unsigned char *commitment,
```
LeoComandini, line 225: Why is `commitment` needed? can't we compute it from `value` and `vbf`?

include/wally_elements.h
```diff
@@ -212,6 +300,46 @@ WALLY_CORE_API int wally_asset_surjectionproof(
     size_t len,
     size_t *written);
 
+/**
+ * Generate an explicit transaction output asset surjection proof.
```
LeoComandini, line 304: same as above

include/wally_elements.h
```diff
@@ -212,6 +300,46 @@ WALLY_CORE_API int wally_asset_surjectionproof(
...
+ *
+ * :param output_asset: asset id for the output.
+ * :param output_asset_len: Length of ``asset``. Must be ``ASSET_TAG_LEN``.
+ * :param output_abf: Asset blinding factor for the output. Generated randomly for each output.
+ * :param output_abf_len: Length of ``output_abf``. Must be ``BLINDING_FACTOR_LEN``.
+ * :param output_generator: Asset generator from `wally_asset_generator_from_bytes`.
```
LeoComandini, line 310: As above, can't we compute the `output_generator` from `output_asset` and `output_abf`?

src/ecdh.c
```diff
@@ -19,13 +19,12 @@ int wally_ecdh(const unsigned char *pub_key, size_t pub_key_len,
...
 
-    if (!pubkey_parse(&pub, pub_key, pub_key_len) ||
-        !seckey_verify(priv_key)) {
+    wally_clear(bytes_out, len);
+    if (!pubkey_parse(&pub, pub_key, pub_key_len))
         ret = WALLY_EINVAL;
-    } else if (!secp256k1_ecdh(ctx, bytes_out, &pub, priv_key, NULL, NULL)) {
-        wally_clear(bytes_out, len);
+    else if (!seckey_verify(priv_key) ||
+             !secp256k1_ecdh(ctx, bytes_out, &pub, priv_key, NULL, NULL))
```
LeoComandini, line 26: why don't we clear `bytes_out` if `secp256k1_ecdh` failed?

src/elements.c
```diff
@@ -235,6 +295,40 @@ int wally_asset_rangeproof(uint64_t value,
...
+
+    if (!rangeproof || rangeproof_len > ASSET_EXPLICIT_RANGEPROOF_MAX_LEN ||
+        parse_commitment(ctx, commitment, commitment_len, &commit) != WALLY_OK ||
+        parse_generator(ctx, generator, generator_len, &gen) != WALLY_OK)
+        return WALLY_EINVAL;
+
+    if (!secp256k1_rangeproof_verify(ctx, &min_v, &max_v, &commit,
+                                     rangeproof, rangeproof_len, NULL, 0, &gen))
+        return WALLY_EINVAL;
+    return value == min_v ? WALLY_OK : WALLY_EINVAL;
```
LeoComandini, line 329: I think we should check `value == max_v` too

src/test/test_elements.py
```diff
@@ -69,6 +80,120 @@ def test_asset_unblind_with_nonce(self):
...
+                                             output_abf, output_abf_len,
+                                             output_generator, output_generator_len,
+                                             explicit_sjproof, explicit_sjproof_len)
+        self.assertEqual(ret, WALLY_OK)
+
+        # explicit_surjectionproof_verify
+        ret = wally_explicit_surjectionproof_verify(explicit_sjproof, explicit_sjproof_len,
+                                                    UNBLINDED_ASSET, UNBLINDED_ASSET_LEN,
+                                                    output_generator, output_generator_len)
+        self.assertEqual(ret, WALLY_OK)
```
LeoComandini, line 194: Please add a test where `wally_explicit_surjectionproof_verify` does not verify

src/test/test_sign.py
```diff
@@ -290,6 +290,61 @@ def test_s2c(self):
...
+        self.assertEqual(sub(scalar, zero), (WALLY_OK, scalar_hex))
+        self.assertEqual(sub(scalar, scalar), (WALLY_OK, zero_hex))
+
+        self.assertEqual(mul(None, zero), (WALLY_EINVAL, zero_hex))
+        self.assertEqual(mul(zero, None), (WALLY_EINVAL, zero_hex))
+        self.assertEqual(mul(zero, bad), (WALLY_ERROR, zero_hex))
+        self.assertEqual(mul(bad, zero), (WALLY_ERROR, zero_hex))
+        self.assertEqual(mul(zero, zero), (WALLY_OK, zero_hex))
+        self.assertEqual(mul(zero, scalar), (WALLY_OK, zero_hex))
+        self.assertEqual(mul(zero, negative), (WALLY_OK, zero_hex))
```
LeoComandini, line 347: Please add some tests for `wally_ec_xonly_verify` too.

src/test/test_psbt.py
```diff
@@ -1,185 +1,291 @@
...
+            self.assertEqual(WALLY_EINVAL, wally_psbt_from_base64(*args))
+
+        self.assertEqual(WALLY_OK, wally_psbt_from_base64(JSON['valid'][0]['psbt'], psbt))
+
+        # psbt_clone_alloc
+        clone = pointer(wally_psbt())
+        for args in [(None, 0x0, clone), # NULL src
+                     (psbt, 0x1, clone), # Invalid flags
+                     (psbt, 0x0, None)]: # NULL dest
+            self.assertEqual(WALLY_EINVAL, wally_psbt_clone_alloc(*args))
```
LeoComandini, line 288: Please add at least a basic test for `wally_psbt_blind`

LeoComandini:
Did a very light review given the huge diff, focused on the elements-related changes.
Added few comments above.

include/wally_crypto.h
jgriffiths: Zero is allowed, text is not (zero or within group order). I'll reword, cheers.

include/wally_elements.h
jgriffiths: It's the max length, zero values can be slightly shorter. So these checks are correct: the first requires the caller pass a sufficient buffer for the largest possible explicit proof, and the second ensures that the passed proof is not trivially invalid (e.g. a full rangeproof).

include/wally_elements.h
jgriffiths: Updated docs, left the name to keep what little consistency there currently is.

include/wally_elements.h
jgriffiths: This is  a library convention to reduce header span, the API docs link to the `with_nonce` variant which documents a superset of the args taken.

include/wally_elements.h
jgriffiths: Removed output ref.

I don't think the wally docs are the place for further documentation, in line with the other function docs it is expected that the caller understands what/wht they need these primitives.

include/wally_elements.h
jgriffiths: 2 reasons:

(1) consistency with the arguments to the non-explicit call.

(2) In common usage you need to compute the commitment for other pruposes anyway, e.g. in PSET we set the commitment, full range proof and the explicit value proof at the same time. computing it 3 times is wasteful.

include/wally_elements.h
jgriffiths: as above

src/ecdh.c
jgriffiths: This call can't fail with a validated pub/privkey which we now check before calling it.

src/elements.c
jgriffiths: This follows the elements `blindpsbt` logic which only checks min value.

src/test/test_sign.py
jgriffiths: Done

src/test/test_psbt.py
jgriffiths: There are no test vectors for this call yet, and the test setup is non-trivial, so a meaningful value test is not possible in a reasonable time frame IMO. If I am granted further time to work on this I will add role tests reconciled with Elements. In the meantime the external users like https://github.com/jgriffiths/wally_swap_test will have to suffice.

JamieDriver:
I'm happy to call it a utack and iterate/fix-forward as required.  (https://github.com/ElementsProject/libwally-core/commit/e1fc3cd4a3e0947d017ad323126b5413538b4710)

Sjors:
> @Sjors Would you mind taking a look at these changes (at least, the interface changes for PSBT v0 in CHANGES.md)?
> 
> Please let me know if you find any issues or would like anything changed, thanks.

Oops, I missed this :-)

When trying to bump to v0.8.6 in LibWallySwift, something did break and I didn't find CHANGES.md very clear on what to do:

* `wally_psbt_output.witness_script` no longer exists, since 8f8481a3c509fc4a02425db3839f92594e6ea852; still figuring out how to use the `psbt_fields map` where this moved

jgriffiths:
@Sjors `include/wally_psbt_members.h` has functions to get and set all members through the top-level psbt object, for the sub objects like inputs and outputs there are only setter functions so that you can create/populate one and add it (then use the top level calls to modify it once added).

This interface is because most wrapped languages can't return a reference to sub-objects once they are embedded in a parent (shared ownership semantics don't transfer well, and taking/returning sub-objects by value is very inefficient).

I'm not against adding getters for the sub-objects for non-swig wrapped language use although it may be a while before I have time to so so.

Sjors:
I ended up using `wally_psbt_get_output_witness_script(_len)` (might be good to add them to the docs). It's not too bad. Although it's slightly inelegant to have to pass `wally_psbt` around, it's probably not a performance issue on a typical iOs device.

PR #337, opened 2022-10-09
src/Makefile.am: split unit tests into separate targets by language

whitslack:
In the typical course of packaging, libwallycore's SWIG Python module may be built for multiple Python interpreters in parallel. Ideally each such resulting module should be tested. However, libwallycore's `Makefile` currently runs all unit tests for all language bindings under a single target. This means, if the Java or JavaScript bindings are enabled, then the unit tests for those other bindings have to be run once for each Python module to be tested. This is redundant and easily avoidable.

Split up `check-local` into four new sub-targets:

  * `check-libwallycore` (`src/test/*.py`)  
    These unit tests are written in Python but do not use (or test!) the wallycore Python module. Indeed, they do not require SWIG even to be installed, as they use Python's ctypes module to link dynamically with libwallycore at run time. They are run for check-local iff tests are enabled and any working Python interpreter is found, regardless of the setting of `--enable-swig-python`.

  * `check-swig-python` (`src/swig_python/contrib/*.py`)  
    These are the unit tests for the `wallycore` Python module. They are run for `check-local` iff `--enable-tests` and `--enable-swig-python`.

  * `check-swig-java` (`src/swig_java/src/com/blockstream/test/*.java`)  
    These are the unit tests for the `wallycore` Java library. They are run for `check-local` iff `--enable-tests` and `--enable-swig-java`.

  * `check-js-wrappers` (`src/wrap_js/test/*.js`)  
    These are the unit tests for the `wallycore` JavaScript wrappers. They are run for `check-local` iff `--enable-tests` and `--enable-js-wrappers`.

This commit relaxes the criteria for `RUN_PYTHON_TESTS` so that it now is enabled whenever any working Python interpreter is found. This implies that `make check` will perform a more thorough check of libwallycore even if the Python module wasn't built, so long as a working Python interpreter is found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -281,7 +281,7 @@ AC_ARG_ENABLE(swig-python,
     [swig_python=$enableval], [swig_python=no])
 AM_CONDITIONAL([USE_SWIG_PYTHON], [test "x$swig_python" == "xyes"])
 
-AM_CONDITIONAL([RUN_PYTHON_TESTS], [test "$PYTHON" != "" -a "x$pythonexists" == "xyes" -a "x$swig_python" == "xyes"])
+AM_CONDITIONAL([RUN_PYTHON_TESTS], [test "$PYTHON" != "" -a "x$pythonexists" == "xyes"])
 
 if test "x$swig_python" == "xyes"; then
     if test "x$pythonexists" != "xyes"; then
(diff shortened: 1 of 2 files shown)


jgriffiths:
@whitslack Good idea, thanks for the patch!

Note I modified it slightly to prevent warnings about duplicated rules.

whitslack:
@jgriffiths: Hmm, I didn't see any warnings. I thought it was valid to add more dependencies to a previously declared target by writing an empty rule. Make should only warn if you try to define more than one recipe for any given target. Anyway, what you did works, though it does mean that, for example, `check-swig-java` will be defined (to do nothing) even if libwally-core is configured with `--disable-swig-java`, whereas the way I did it, you'd get a "no rule to make check-swig-java" if you tried to invoke it explicitly in that circumstance. No big deal, I guess.

PR #338, opened 2022-10-09
m4/ax_jni_include_dir: don't cache result

whitslack:
The `AX_JNI_INCLUDE_DIR` macro was violating the first rule of `AC_CACHE_CHECK`: ["The _commands-to-set-it_ _must have no side effects_ except for setting the variable _cache-id_."][1] Specifically, `JNI_INCLUDE_DIRS` was being modified within the code block that was only supposed to be setting the value of `ac_cv_jni_header_path`. This was causing `JNI_INCLUDE_DIRS` to be incorrect/incomplete when `ac_cv_jni_header_path` was initialized from an Autoconf cache file.

Since the tests to find the JNI header paths are so trivial, there is nothing to be gained by caching the results. The most straightforward fix is not to cache, and that is the fix provided in this PR.

This PR is based on a [patch](https://savannah.gnu.org/patch/index.php?10284) that I submitted to GNU Savannah, but that patch does not cleanly apply here because libwally-core's local copy is not in sync with the [most recent version](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_jni_include_dir.m4) in the GNU Autoconf Archive.

[1]: https://www.gnu.org/software/autoconf/manual/html_node/Caching-Results.html

### DIFF
--- a/tools/build-aux/m4/ax_jni_include_dir.m4
+++ b/tools/build-aux/m4/ax_jni_include_dir.m4
@@ -32,6 +32,10 @@
 #
 #   - at the configure level, setenv JAVAC
 #
+#   This macro depends on AC_CANONICAL_HOST which requires that config.guess
+#   and config.sub be distributed along with the source code.  See autoconf
+#   manual for details.
+#
 #   Note: This macro can work with the autoconf M4 macros for Java programs.
 #   This particular macro is not part of the original set of macros.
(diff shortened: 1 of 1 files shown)


jgriffiths:
Merged, thanks!

PR #339, opened 2022-10-09
src/Makefile.am: add missing headers to install

whitslack:
It looks like `Makefile.am` was neglected when `wally_map.h` and `wally_psbt_members.h` were introduced. Without the fix in this PR, those headers don't get installed.


jgriffiths:
Merged, thanks!

PR #341, opened 2022-10-10
Bip32 updates

jgriffiths:
Includes the bip32 mixed-hardened/non-hardened derivation fixes and upper case bip32 path changes from the descriptor PR, plus adds a couple more test cases from @JamieDriver

### DIFF
--- a/src/test/test_bip32.py
+++ b/src/test/test_bip32.py
@@ -399,9 +399,9 @@ class BIP32Tests(unittest.TestCase):
             self.compare_keys(master, expected, flags)
 
         derived = master
-        for path, i in [('m/0H', 0x80000000),
-                        ('m/0H/1', 1),
-                        ('m/0H/1/2H', 0x80000002)]:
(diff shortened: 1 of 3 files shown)


src/test/test_bip32.py
```diff
@@ -533,33 +539,54 @@ def get_paths(path):
...
+        cases = [('m/1/*',     55),
+                 ('m/*',       55),
+                 ('m/1/*/1',   55),
+                 ('m/1h/*',    55),
+                 ('m/1h/*/1',  55),
+                 ('m/1h/*/1h', 55),
+                 ('m/1/*h',    55),
+                 ('m/*h',      55),
+                 ('m/1/*h/1',  55),
+                 ('m/1/*h/1h', 55)]
```
JamieDriver, line 573: Could we add here something like: `m/*h/1h/1h/1/1`, as that covers the bip44/48/84 schemes - common singlesig use cases.  Cheers.

PR #343, opened 2022-11-13
Issuance fixes, map/keypath/PSBT helpers

jgriffiths:


src/map.c
```diff
@@ -560,6 +560,68 @@ int wally_map_keypath_get_item_fingerprint(const struct wally_map *map_in, size_
...
 
+int wally_keypath_get_path_len(const unsigned char *val, size_t val_len,
+                               size_t *written)
+{
+    if (written)
+        *written = 0;
+    if (!val || val_len < BIP32_KEY_FINGERPRINT_LEN || val_len % sizeof(uint32_t))
+        return WALLY_EINVAL;
+    if (val_len > BIP32_KEY_FINGERPRINT_LEN)
+        *written = (val_len / sizeof(uint32_t)) - 1;
```
JamieDriver, line 571: nit:  this code assumes that BIP32_KEY_FINGERPRINT_LEN == sizeof(uint32_t)  [which ofc it does].
But it might be better to not ?
eg:
`if (!val || val_len < BIP32_KEY_FINGERPRINT_LEN || (val_len - BIP32_KEY_FINGERPRINT_LEN) % sizeof(uint32_t))`
and:
`*written = (val_len - BIP32_KEY_FINGERPRINT_LEN) / sizeof(uint32_t));`
?

JamieDriver:
Eyeballed, lgtm.  
I've also updated local wip code to use the new map functions and 'get_best_utxo()' and all works well.
(As my code also uses the custom-key-from-seed fn, my unit tests should also cover the new hdkey partial-init fn).
On these grounds, ack 1b8a248
[ caveat, not familiar with the pset-specific issuance changes ]

src/map.c
jgriffiths: Agreed, consistency is better. I've added two small helpers to do these calcs and changed the code to use them, thanks.

JamieDriver:
re-tested, all good.  ack ee0a92a

PR #349, opened 2022-11-23
Final candidate 0.8.7 changes

jgriffiths:


Sjors:
I'm pretty close to getting 0.8.6 to work with libwally-swift. I'll let you know if I run into any issues that might be nice for 0.8.7

jgriffiths:
> (done)

Thanks @Sjors - 0.8.7 is coming very shortly and has some map function helpers which should allow you to wrap map functions more nicely (for example, in python we can convert to and from python dicts and maps trvially now).

0.8.8 will also be out pretty quickly after that, and will include updates to the signing interface to allow bespoke signing without having to re-implement all of the logic in the current `psbt_sign` call. As an example, you'll be able to sign individual inputs, and sign directly using a bip32 hdkey without having to find and derive matching private keys yourself.

jgriffiths:
@k-matsuzawa @ko-matsu I'd be interested in your feedback on the new WASM-based JS/Typescript wrapper.

The intention is to replace the existing wrappers with the new ones, since the new ones cover the entire wally API now, and are auto-generated instead of having to be manually updated. I expect to remove the old wrappers in release 0.8.8, please let me know if you are able to switch over and/or whether this will cause any issues for you, thanks.

k-matsuzawa:
@jgriffiths Thanks! This is so good! I will switch to the new JS/Typescript wrapper.

CHANGES.md
```diff
@@ -1,5 +1,22 @@
...
+- psbt_extract now takes an extra flags parameter. The Java and Python
+  wrappers default this to zero for backwards compatibility.
+- bip39_mnemonic_to_seed is no longer available for SWIG-wrapped languages.
+  Callers should use bip39_mnemonic_to_seed512 instead.
+- Python: Some calls (e.g. bip39_mnemonic_to_seed512, aes, and aes_cbc) used
+  to take output buffers to fill and return the number of bytes written. These
+  calls now automatically create and fill thier buffer, returning it as the
+  only return value. The number of bytes written can be determined by using
+  len() on the returned output buffer.
+
```
LeoComandini, line 19: I would add
```suggestion
Remove support for Python 2.x.
```
or something similar

src/test/test_map.py
```diff
@@ -298,6 +301,32 @@ def test_keypath_map(self):
...
 
+        # BIP32 lookup
+        bip32_invalid = [
+            (None, 0, derived), # Null map
+            (m, 0, None),       # Null key
+        ]
+        # - compressed
+        ret = wally_map_add(m, pub_key, pub_key_len, kp_path, len(kp_path))
+        self.assertEqual(ret, WALLY_OK)
+        ret = wally_map_find_bip32_public_key_from(m, 0, derived)
```
LeoComandini, line 312: do you think it's worth to add a couple of tests for `wally_map_find_from`?

LeoComandini:
utACK d868fb1c898dbf746ca69bea89e11c9699208edd, code review, haven't reviewed the js changes, tested locally, did a couple of smoke tests.

PR #350, opened 2022-11-24
wasm: Fix length functions

shesek:


shesek:
Cherry picked into https://github.com/elementsproject/libwally-core/tree/candidate_87.

PR #352, opened 2022-12-02
wasm: Fix Webpack compatibility for browser use

shesek:
Updating emsdk to v3.1.27 was necessary for https://github.com/emscripten-core/emscripten/pull/17915, which allows using the same WASM build for both browser and nodejs envs.


jgriffiths:
Merged/squashed with updates, thanks!

PR #369, opened 2023-02-13
descriptor: Implement wallet-policies BIP

jgriffiths:
See https://github.com/bitcoin/bips/pull/1389 for details of wallet policies.

Note wally implements the `/*` extension to support non-BIP44 wallets such as Blockstream Green. Additionally, we do not limit allowable policies to the whitelist in the linked BIP. We allow all valid descriptors where the solved cardinality is either 1 or 2.

Includes some extra functionality that is useful for validating descriptors/policies, primarily the ability to iterate and query the keys in the parsed expression.


include/wally_descriptor.h
```diff
@@ -15,6 +15,7 @@ struct wally_descriptor;
 #define WALLY_MINISCRIPT_TAPSCRIPT        0x01 /** Tapscript, use x-only pubkeys */
 #define WALLY_MINISCRIPT_ONLY             0x02 /** Only allow miniscript (not descriptor) expressions */
 #define WALLY_MINISCRIPT_REQUIRE_CHECKSUM 0x04 /** Require a checksum to be present */
+#define WALLY_MINISCRIPT_POLICY           0x08 /** Only allow policy @n variable substitution */
```
bigspider, line 18: In the context of miniscript, the word _policy_ already has a [specific meaning](https://bitcoin.sipa.be/miniscript/#:~:text=Policy%20to%20Miniscript%20compiler), which doesn't quite match the meaning of my _wallet policy_ proposal (which rather builds on top of descriptors − which might or might not contain miniscript within `wsh` or `tr`).

Pointing that out as I suspect the name `WALLY_MINISCRIPT_POLICY` might be confusing.

jgriffiths: Agreed, I will wait for the resolution of the question on your BIP before merging, and have thrown `wallet patterns` into the ring as a contender.

PR #371, opened 2023-02-24
psbt: verify any provided sighash matches sigs when parsing

jgriffiths:
Core doesn't verify this on parsing, BIP-0174 states only that finalizers must fail to finalize where sighashes don't match.

Block this from happening when parsing to avoid processing an invalid PSBT (that a buggy finalizer may not catch), and to avoid needing to check the sighash in the finalization code for each supported input type. Note we already prevent signatures being added that don't match.


instagibbs:
ACK 5d2a281717d9e193d17f2609686d637bdc212613

Checked the test examples had mismatching sighashes/signatures (NONE vs ALL for each)

Checked `wally_psbt_input_set_sighash` logic to make sure it will fail if signatures don't match for that input

PR #373, opened 2023-03-01
Psbt updates

jgriffiths:
Primarily makes finalization more flexible, and adds flags to support buggy behavior seen by other implementations, which external software (e.g. gdk) already relies on.

Some minor fixes and also allows detecting CSV scripts in preparation for supporting finalizing them.

**NOTE:** This PR contains ABI changes as per `CHANGES.md`.


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Version 0.8.8
 - scriptpubkey_csv_2of3_then_2_from_bytes has been removed from the API.
+- psbt: psbt_extract now takes an extra flags parameter. The Java and Python
```
instagibbs, line 5: this seems out of place? commit is about `psbt_finalize`?

instagibbs (approved):
ACK https://github.com/ElementsProject/libwally-core/pull/373/commits/609a41465b28589761cc3a5f7bf46ade5f5071ff

Every change is straight forward. CHANGES.md likely a typo though.

CHANGES.md
jgriffiths: Yes good catch, c'n'p from a previous API change, will update.

PR #374, opened 2023-03-10
Rename variable to clarify asset commitments and generator

philippem:
The example is very helpful for explaining CT in Liquid as well as the API in libwally.

This MR clarifies that an asset commitment is stored in a CT transaction output (not an asset), and that an asset commitment is also a generator.


jgriffiths:
Merged, thanks!

PR #375, opened 2023-03-14
add support for mbedtls sha245 and sha512 for esp-idf with a compile flag

greenaddress:


jgriffiths:
Rebased and merged with configure support, thanks!

PR #379, opened 2023-03-21
Support schnorr sigs in wally_ec_sig_from_bytes(_len), wally_ec_sig_verify

instagibbs:
Test vectors taken from https://github.com/bitcoin/bips/blob/master/bip-0340/test-vectors.csv


src/sign.c
```diff
@@ -304,8 +302,6 @@ int wally_ec_sig_from_bytes_len(const unsigned char *priv_key, size_t priv_key_l
     if (flags & EC_FLAG_SCHNORR) {
         if (flags & EC_FLAG_RECOVERABLE)
```
jgriffiths, line 303: For schnorr we should reject `EC_FLAG_GRIND_R` as well I think.

src/sign.c
```diff
@@ -329,10 +325,14 @@ int wally_ec_sig_from_bytes(const unsigned char *priv_key, size_t priv_key_len,
...
     if (flags & EC_FLAG_SCHNORR) {
+        secp256k1_keypair keypair;
         if (flags & EC_FLAG_RECOVERABLE)
             return WALLY_EINVAL; /* Only ECDSA is supported for recoverable sigs */
-        /* FIXME: Implement taproot schnorr sigs */
-        return WALLY_ERROR;
+        if (secp256k1_keypair_create(ctx, &keypair, priv_key) != 1)
+            return WALLY_ERROR;
+        if (secp256k1_schnorrsig_sign32(ctx, bytes_out, bytes, &keypair, /* aux_rand32 */ NULL) != 1)
+            return WALLY_ERROR;
```
jgriffiths, line 338: `keypair` needs to be cleared for both success and failure cases before returning here.

src/sign.c
```diff
@@ -373,30 +373,35 @@ int wally_ec_sig_verify(const unsigned char *pub_key, size_t pub_key_len,
...
     secp256k1_ecdsa_signature sig_secp;
     const secp256k1_context *ctx = secp_ctx();
     bool ok;
 
-    if (!pub_key || pub_key_len != EC_PUBLIC_KEY_LEN ||
-        !bytes || bytes_len != EC_MESSAGE_HASH_LEN ||
+    if (!pub_key || (pub_key_len != EC_PUBLIC_KEY_LEN &&
+        pub_key_len != EC_XONLY_PUBLIC_KEY_LEN))
+        return WALLY_EINVAL;
+
```
jgriffiths, line 404: You check these lengths below for each case as well so you can remove this whole `if` and just add pub_key to the checks immediately below I reckon.

src/sign.c
```diff
@@ -373,30 +373,35 @@ int wally_ec_sig_verify(const unsigned char *pub_key, size_t pub_key_len,
...
-#if 0 /*FIXME: Schnorr is unavailable in secp for now*/
-        ok = ok && secp256k1_schnorr_verify(ctx, sig, bytes, &pub);
-#else
-        ok = false;
-#endif
-    else
+    if (flags & EC_FLAG_SCHNORR) {
+        if (pub_key_len != EC_XONLY_PUBLIC_KEY_LEN)
+            return WALLY_EINVAL;
+		ok = xpubkey_parse(&xonly_pub, pub_key, pub_key_len);
```
jgriffiths: - Misleading indenting.
-  `xpubkey_parse` will take either a full or an xonly pubkey, so there is no harm in removing the length check and allowing the caller to verify with a normal compressed pubkey IMO. The fewer places that force the distinction on the caller externally, the better.

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
         self.assertEqual(mul(zero, scalar), (WALLY_OK, zero_hex))
         self.assertEqual(mul(zero, negative), (WALLY_OK, zero_hex))
 
+    def test_bip340_sigs(self):
+
+        made_sig = False
+
+        for case in self.get_bip340_sign_cases():
+            priv_key = make_cbuffer(case["priv_key"])
```
jgriffiths, line 394: I prefer e.g. `priv_key, priv_key_len = make_cbuffer(case['priv_key'])` and then remove the `[0]` and `[1]` subscripting below.

Please note the use of `'` instead of `"` for strings (except autodoc lines) in the python code, this for for ease of grepping.

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+            pub_key = make_cbuffer(case["pub_key"])
+            aux_rand = make_cbuffer(case["aux_rand"])
+            msg = make_cbuffer(case["msg"])
+            sig = make_cbuffer(case["sig"])
+            verification = case["verification"]
+
+            sig_out = self.cbufferize(['00' * 64])[0]
+
+            # We don't expose aux_rand, so only test default value for now
+            if priv_key[1] == 32 and aux_rand[0] == bytes(32):
```
jgriffiths: Please fix the existing typo `EX_PRIV_KEY_LEN` to `EC_PRIV_KEY_LEN` and use that in this test instead of `32` (the aux rand length is ok as 32 for now as we don't have a constant for it).

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+                self.assertEqual(sig[0], sig_out)
+                made_sig = True
+
+            if priv_key[1] == 32:
+                ret = wally_ec_sig_from_bytes_len(priv_key[0], priv_key[1],
+                                                  msg[0], msg[1], FLAG_SCHNORR)
+                self.assertEqual(ret, (WALLY_OK, EC_SIGNATURE_LEN))
+
+                # Wrong size privkey
+                ret = wally_ec_sig_from_bytes_len(priv_key[0][:-1], priv_key[1] - 1,
```
jgriffiths, line 415: you should remove the `[:-1]` from the first argument here, only the pointer is being passed so there's no need to take a slice.

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+                                                  msg[0], msg[1], FLAG_SCHNORR)
+                self.assertEqual(ret, (WALLY_EINVAL, 0))
+
+                # TODO support wally_ec_public_key_from_private_key with bip340 keys to check priv_key -> pub_key
+
+            self.assertEqual(verification, wally_ec_sig_verify(pub_key[0], pub_key[1], msg[0], msg[1], FLAG_SCHNORR, sig[0], sig[1]) == WALLY_OK)
```
jgriffiths, line 421: - Perhaps rename `verification` to `is_valid` here and in the json?
- Try to keep lines to 100 chars or less, here you can call `ret = wally_ec_sig_verify(...)` and then `self.assertEqual(is_valid, ret == WALLY_OK)` below.

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+
+            # Invalid flag
```
jgriffiths, line 423: `# Wrong signature type` is more correct, the flag is valid for this call, its just not what the data represents.

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+            self.assertEqual(WALLY_EINVAL, wally_ec_sig_verify(pub_key[0], pub_key[1], msg[0], msg[1], FLAG_ECDSA, sig[0], sig[1]))
+
+            # Wrong size pubkey
+            self.assertEqual(WALLY_EINVAL, wally_ec_sig_verify(pub_key[0][:-1], pub_key[1] - 1, msg[0], msg[1], FLAG_SCHNORR, sig[0], sig[1]))
```
jgriffiths, line 427: as above, remove `[:-1]`

src/test/test_sign.py
```diff
@@ -377,5 +382,49 @@ def add_by_final_vbf(a_hex, b_hex):
...
+
+        assert made_sig
```
jgriffiths, line 429: To ensure you tested something, count the number of cases in a var like `num_tests_run` and then `self.assertGreater(0, num_tests_run)` here.

src/sign.c
instagibbs: done, I think

src/test/test_sign.py
instagibbs: done

src/test/test_sign.py
instagibbs: done

src/test/test_sign.py
instagibbs: done

src/test/test_sign.py
instagibbs: done

src/test/test_sign.py
instagibbs: done

jgriffiths:
Merged, many thanks for the responsive review updates.

PR #381, opened 2023-03-28
sign: support aux data for ECDSA, docs fixes, fix secp version

jgriffiths:
secp was incorrectly reverted when merging the previous commit.

This enables caller use of the aux_rand secp feature for ECDSA, which can be used e.g. for manual grinding.


instagibbs:
ACK https://github.com/ElementsProject/libwally-core/pull/381/commits/a28bcfd0a6c8d50a71542e0f5b97e698ff244a2b

PR #390, opened 2023-04-19
Add taproot keypath setters for PSBT

instagibbs:
Opening this to see if I'm on the right track. Needs tests.


include/wally_map.h
```diff
@@ -462,6 +462,34 @@ WALLY_CORE_API int wally_map_keypath_add(
...
 
+/**
+ * Convert and add a public key, leaves and path to taproot leafhash/keypath maps.
+ *
+ * :param hashes_map_in: The taproot_leaf_hashes map to add to.
+ * :param paths_map_in: The taproot_leaf_paths map to add to.
+ * :param pub_key: The public key or extended public key to add.
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `BIP32_SERIALIZED_LEN`
+ *|    for an extended bip32 public key, or `EC_XONLY_PUBLIC_KEY_LEN`
+ *|    for a public key.
```
jgriffiths, line 473: I think only EC_XONLY_PUBLIC_KEY_LEN is valid?

include/wally_map.h
```diff
@@ -462,6 +462,34 @@ WALLY_CORE_API int wally_map_keypath_add(
...
+ * :param child_path: The BIP32 derivation path for the pubkey.
+ * :param child_path_len: The number of items in ``child_path``.
+ */
+WALLY_CORE_API int wally_map_tap_keypath_add(
+    struct wally_map *hashes_map_in,
+    struct wally_map *paths_map_in,
+    const unsigned char *pub_key,
+    size_t pub_key_len,
+    const unsigned char *tapleaf_hashes,
+    size_t tapleaf_hashes_len,
```
jgriffiths, line 487: I think `leaf_hashes` or 'merkle_hashes` is fine here - when wally exposes merkle tree ops we can reuse the same parameter name for consistency.

src/map.c
```diff
@@ -664,6 +664,49 @@ int wally_map_keypath_add(struct wally_map *map_in,
...
+
+    if (!hashes_map_in || !paths_map_in || !fingerprint || fingerprint_len != BIP32_KEY_FINGERPRINT_LEN ||
+        BYTES_INVALID(path, path_len) || pub_key_len != EC_XONLY_PUBLIC_KEY_LEN)
+        return WALLY_EINVAL;
+
+	/* TODO add hashes_map_in hash verification */
+    if (paths_map_in->verify_fn != wally_keypath_xonly_public_key_verify)
+        return WALLY_EINVAL; /* Not a taproot keypath map */
+
+	/* TODO take the tapleaf hashes value instead of copy */
```
jgriffiths, line 686: You can remove this, wally always needs internal ownership of held data and so will always take a copy.

src/map.c
```diff
@@ -664,6 +664,49 @@ int wally_map_keypath_add(struct wally_map *map_in,
     return ret;
 }
 
+int wally_map_tap_keypath_add(struct wally_map *hashes_map_in,
+                          struct wally_map *paths_map_in,
+                          const unsigned char *pub_key, size_t pub_key_len,
+                          const unsigned char *tapleaf_hashes, size_t tapleaf_hashes_len,
+                          const unsigned char *fingerprint, size_t fingerprint_len,
+                          const uint32_t *path, size_t path_len)
```
jgriffiths, line 672: At the `wally_map` interface level, I think its better to split this into two functions, and remove `tap` from the name for potential future re-use of maps holding merkle paths.

IIUC though, the final part of this impl is just `wally_map_keypath_add` on `paths_map_in`. I'd leave that as a separate call since its already implemented/exposed.

This function then would become `wally_map_merkle_path add` which just adds the merkle path keyed by the pubkey to `hashes_map_in`. If `hashes_map_in` has a verify function that only allows values from (0..N) * SHA256_LEN in length (which for psbt it does, see `map_leaf_hashes_verify`) then I think it also becomes redundant at the `wally_map` interface level.

For psbt in contrast, since (1) you want to add entries to both maps in tandem, and (2) this is taproot specific then implementing `wally_psbt_[input|output]_taproot_path_add` taking both the hashes and path info makes sense.

src/psbt.c
```diff
@@ -194,6 +194,22 @@ static const struct wally_tx_output *utxo_from_input(const struct wally_psbt *ps
...
+                               const unsigned char *fingerprint, size_t fingerprint_len, \
+                               const uint32_t *child_path, size_t child_path_len) { \
+        if (!parent) return WALLY_EINVAL; \
+        return wally_map_tap_keypath_add(&parent->taproot_leaf_hashes, \
+                                         &parent->taproot_leaf_paths, \
+                                         pub_key, pub_key_len, \
+                                         fingerprint, fingerprint_len, \
+                                         tapleaf_hashes, tapleaf_hashes_len, \
+                                         child_path, child_path_len); \
+    }
```
jgriffiths, line 211: As per other comment, I would not do this at the map interface level. You could have a local static function with this signature that this macro calls, but I suspect that would be just more code than its worth vs calling the 2 map functions separately.

include/wally_psbt.h
```diff
@@ -303,6 +303,30 @@ WALLY_CORE_API int wally_psbt_input_keypath_add(
...
+ * :param pub_key: The pubkey to add.
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `EC_XONLY_PUBLIC_KEY_LEN`.
+ * :param tapleaf_hashes: Series of 32-byte leaf hashes.
+ * :param tapleaf_hashes_len: Length of ``tapleaf_hashes`` in bytes.
+ * :param fingerprint: The master key fingerprint for the pubkey.
+ * :param fingerprint_len: Length of ``fingerprint`` in bytes. Must be `BIP32_KEY_FINGERPRINT_LEN`.
+ * :param child_path: The BIP32 derivation path for the pubkey.
+ * :param child_path_len: The number of items in ``child_path``.
+ */
+WALLY_CORE_API int wally_psbt_input_tap_keypath_add(
```
jgriffiths, line 319: Use `taproot` rather than `tap` to match `wally_tx_get_btc_taproot_signature_hash`.

src/map.c
instagibbs: ah ok, `wally_map_keypath_add` seems to clear out the caller's field, I'll remove

src/map.c
instagibbs: > IIUC though, the final part of this impl is just wally_map_keypath_add on paths_map_in. I'd leave that as a separate call since its already implemented/exposed.

The key(badumsh) difference here is that `wally_keypath_public_key_verify` doesn't allow xonly keys. Should I relax `wally_map_keypath_add` to accept `wally_keypath_xonly_public_key_verify` as a `verify_fn` for the map, then reuse it?

src/psbt.c
instagibbs: to be clear, do you mean I should modify the macro to call `wally_map_merkle_path_add` and `wally_map_keypath_add` since this `wally_map_tap_keypath_add` would be going away?

include/wally_psbt.h
```diff
@@ -1297,6 +1321,30 @@ WALLY_CORE_API int wally_psbt_output_keypath_add(
...
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `EC_XONLY_PUBLIC_KEY_LEN`.
+ * :param tapleaf_hashes: Series of 32-byte leaf hashes.
+ * :param tapleaf_hashes_len: Length of ``tapleaf_hashes`` in bytes.
+ * :param fingerprint: The master key fingerprint for the pubkey.
+ * :param fingerprint_len: Length of ``fingerprint`` in bytes. Must be `BIP32_KEY_FINGERPRINT_LEN`.
+ * :param child_path: The BIP32 derivation path for the pubkey.
+ * :param child_path_len: The number of items in ``child_path``.
+ */
+WALLY_CORE_API int wally_psbt_output_tap_keypath_add(
+    struct wally_psbt_input *input,
```
instagibbs, line 1338: oops, this needs to be output

include/wally_map.h
instagibbs: done

src/map.c
instagibbs: done

src/psbt.c
instagibbs: done

include/wally_psbt.h
instagibbs: done

src/map.c
jgriffiths: It makes a copy which it swaps in, so the action is atomic.

include/wally_map.h
```diff
@@ -420,6 +420,20 @@ WALLY_CORE_API int wally_keypath_xonly_public_key_verify(
     const unsigned char *val,
     size_t val_len);
 
+/**
+ * Verify a PSBT merkle path keyed by an x-only public key.
```
jgriffiths, line 424: You can remove PSBT from this since it works with any merkle path.

include/wally_map.h
```diff
@@ -446,8 +460,8 @@ WALLY_CORE_API int wally_map_keypath_public_key_init_alloc(
  * :param map_in: The keypath map to add to.
  * :param pub_key: The public key or extended public key to add.
  * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `BIP32_SERIALIZED_LEN`
- *|    for an extended bip32 public key, or `EC_PUBLIC_KEY_UNCOMPRESSED_LEN`
- *|    or `EC_PUBLIC_KEY_LEN` for a public key.
+ *|    for an extended bip32 public key, or `EC_PUBLIC_KEY_UNCOMPRESSED_LEN`,
```
jgriffiths, line 463: remove the first `or`

include/wally_map.h
```diff
@@ -462,6 +476,23 @@ WALLY_CORE_API int wally_map_keypath_add(
...
 
+/**
+ * Convert and add a public key and merkle path to map.
+ *
+ * :param map_in: The map to add to.
+ * :param pub_key: The public key or extended public key to add.
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `EC_XONLY_PUBLIC_KEY_LEN`
+ *|    bytes.
+ * :param merkle_hashes: Series of 32-byte leaf hashes.
+ * :param merkle_hashes_len: Length of ``merkle_hashes`` in bytes.
```
jgriffiths, line 487: + 'must be a multiple of SHA256_LEN

include/wally_psbt.h
```diff
@@ -303,6 +303,30 @@ WALLY_CORE_API int wally_psbt_input_keypath_add(
     const uint32_t *child_path,
     size_t child_path_len);
 
+/**
+ * Convert and add a pubkey/keypath to a taproot input.
+ *
```
jgriffiths, line 308: to a *PSBT* input.

include/wally_psbt.h
```diff
@@ -303,6 +303,30 @@ WALLY_CORE_API int wally_psbt_input_keypath_add(
...
+ * :param input: The input to add to.
+ * :param pub_key: The pubkey to add.
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `EC_XONLY_PUBLIC_KEY_LEN`.
+ * :param tapleaf_hashes: Series of 32-byte leaf hashes.
+ * :param tapleaf_hashes_len: Length of ``tapleaf_hashes`` in bytes.
```
jgriffiths, line 313: as above, must be a multiple...

include/wally_psbt.h
```diff
@@ -1297,6 +1321,30 @@ WALLY_CORE_API int wally_psbt_output_keypath_add(
...
+ * :param output: The output to add to.
+ * :param pub_key: The pubkey to add.
+ * :param pub_key_len: Length of ``pub_key`` in bytes. Must be `EC_XONLY_PUBLIC_KEY_LEN`.
+ * :param tapleaf_hashes: Series of 32-byte leaf hashes.
+ * :param tapleaf_hashes_len: Length of ``tapleaf_hashes`` in bytes.
+ * :param fingerprint: The master key fingerprint for the pubkey.
+ * :param fingerprint_len: Length of ``fingerprint`` in bytes. Must be `BIP32_KEY_FINGERPRINT_LEN`.
+ * :param child_path: The BIP32 derivation path for the pubkey.
+ * :param child_path_len: The number of items in ``child_path``.
+ */
```
jgriffiths, line 1336: as above for input comments

src/map.c
```diff
@@ -620,6 +620,20 @@ int wally_keypath_xonly_public_key_verify(const unsigned char *key, size_t key_l
     return WALLY_OK;
 }
 
+int wally_merkle_path_xonly_public_key_verify(const unsigned char *key, size_t key_len,
+                                          const unsigned char *val, size_t val_len)
+{
```
jgriffiths, line 625: nit: formatting

src/map.c
```diff
@@ -620,6 +620,20 @@ int wally_keypath_xonly_public_key_verify(const unsigned char *key, size_t key_l
...
+    struct ext_key extkey;
+
+    if (key_len != EC_XONLY_PUBLIC_KEY_LEN ||
+        keypath_key_verify(key, key_len, &extkey) != WALLY_OK ||
+        extkey.version ||
+        !val ||
+        val_len % SHA256_LEN != 0)
```
jgriffiths, line 632: `val_len` of zero should be failed?

also nit: please bunch up the conditions.

src/map.c
```diff
@@ -664,6 +679,22 @@ int wally_map_keypath_add(struct wally_map *map_in,
     return ret;
 }
 
+int wally_map_merkle_path_add(struct wally_map *map_in,
+                          const unsigned char *pub_key, size_t pub_key_len,
+                          const unsigned char *merkle_hashes, size_t merkle_hashes_len)
+{
+
```
jgriffiths, line 686: bit: blank line, formatting

src/map.c
```diff
@@ -664,6 +679,22 @@ int wally_map_keypath_add(struct wally_map *map_in,
...
+    if (!map_in || pub_key_len != EC_XONLY_PUBLIC_KEY_LEN)
+        return WALLY_EINVAL;
+
+    if (map_in->verify_fn != wally_merkle_path_xonly_public_key_verify)
+        return WALLY_EINVAL; /* Not a taproot keypath map */
+
+    /* Add map for tap leaves */
+    return map_add(map_in, pub_key, pub_key_len,
+                  merkle_hashes_len ? merkle_hashes : NULL, merkle_hashes_len, false, false);
+}
```
jgriffiths, line 696: formatting! also please keep line lengths to less than 100 if possible.

jgriffiths: >merkle_hashes_len ? merkle_hashes : NULL

Its always an error in wally to pass a valid ptr with 0 length or a null ptr with a length. Please check this using `BYTES_INVALID` above and remove the ternary.

src/psbt.c
```diff
@@ -194,6 +194,25 @@ static const struct wally_tx_output *utxo_from_input(const struct wally_psbt *ps
...
+/* Add a taproot keypath to parent structs keypaths member */
+#define ADD_TAP_KEYPATH(PARENT) \
+    int PARENT ## _tap_keypath_add(struct PARENT *parent, \
+                               const unsigned char *pub_key, size_t pub_key_len, \
+                               const unsigned char *tapleaf_hashes, size_t tapleaf_hashes_len, \
+                               const unsigned char *fingerprint, size_t fingerprint_len, \
+                               const uint32_t *child_path, size_t child_path_len) { \
+        int ret; \
+        if (!parent) return WALLY_EINVAL; \
+        ret = wally_map_merkle_path_add(&parent->taproot_leaf_hashes, \
```
jgriffiths, line 206: so to be strictly correct here you need to validate the args for the second call before making the first, to avoid adding to only one map (where possible, operations should be atomic). You can do this with the validation function from `parent->taproot_leaf_paths`.

jgriffiths (commented):
If you address the above and squash I can look at why this isn't building. Please check all formatting, its time consuming to have to point the instances out, thanks.

instagibbs:
Believe I've handled all outstanding comments.

jgriffiths (commented):
LGTM.

Needs tests. I'm happy with just invalid value tests, and perhaps serialization tests with invalid script paths (not a multiple of the hash size, or with a-non-x-only pubkey, for example). The valid value/happy path tests can be part of the psbt script spend tests when you plug in these calls to the psbt logic.

In order to add those tests you'll probably need to expose the psbt level add function, that is as well as `wally_psbt_[input|output]_taproot_keypath_add`, you need `wally_psbt_add_[input|output]_taproot_keypath` which takes a psbt and the index of the input/output. As it stands only non-wrapped languages can make use of the add calls.

src/psbt.c
```diff
@@ -194,6 +194,26 @@ static const struct wally_tx_output *utxo_from_input(const struct wally_psbt *ps
                                      child_path, child_path_len); \
     }
 
+/* Add a taproot keypath to parent structs keypaths member */
+#define ADD_TAP_KEYPATH(PARENT) \
+    int PARENT ## _tap_keypath_add(struct PARENT *parent, \
```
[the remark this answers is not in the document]
instagibbs: ah yep, fixed

src/test/test_psbt.py
```diff
@@ -1,3 +1,4 @@
+import ctypes
```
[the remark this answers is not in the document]
jgriffiths: you don't need the import, see the bip32 test for other examples, but e.g. `foo = (c_uint32 * some_length)()` will make a c array of length `some_length`.

src/psbt.c
```diff
@@ -1436,6 +1460,36 @@ static int psbt_input_from_tx_input(struct wally_psbt *psbt,
...
+                                         size_t pub_key_len,
+                                         const unsigned char *tapleaf_hashes,
+                                         size_t tapleaf_hashes_len,
+                                         const unsigned char *fingerprint,
+                                         size_t fingerprint_len,
+                                         const uint32_t *child_path,
+                                         size_t child_path_len)
+{
+    if (!psbt_is_valid(psbt) || (psbt->version == PSBT_0 && !psbt->tx) ||
+        index > psbt->num_inputs || flags)
```
jgriffiths, line 1475: `>` should be `>=` here. But use the `psbt_get_input`/`psbt_get_output` helpers since they do this checking including the v0 checks for you.

src/psbt.c
```diff
@@ -1436,6 +1460,36 @@ static int psbt_input_from_tx_input(struct wally_psbt *psbt,
...
+        return WALLY_EINVAL;
+
+    if (!psbt_can_modify(psbt, WALLY_PSBT_TXMOD_INPUTS))
+        return WALLY_EINVAL;
+
+
```
jgriffiths, line 1481: you can fold this into the previous if, and remove the excess blank lines.

src/psbt.c
```diff
@@ -1436,6 +1460,36 @@ static int psbt_input_from_tx_input(struct wally_psbt *psbt,
...
+int wally_psbt_add_input_taproot_keypath(struct wally_psbt *psbt,
+                                         uint32_t index, uint32_t flags,
+                                         const unsigned char *pub_key,
+                                         size_t pub_key_len,
+                                         const unsigned char *tapleaf_hashes,
+                                         size_t tapleaf_hashes_len,
+                                         const unsigned char *fingerprint,
+                                         size_t fingerprint_len,
+                                         const uint32_t *child_path,
+                                         size_t child_path_len)
```
jgriffiths, line 1472: for long function names/lots of arguments, you can wrap these args and pair them up e.g:
```
int wally_psbt_add_input_taproot_keypath(
    struct wally_psbt *psbt, uint32_t index, uint32_t flags,
    const unsigned char *pub_key, size_t pub_key_len,
    const unsigned char *tapleaf_hashes, size_t tapleaf_hashes_len,
    const unsigned char *fingerprint, size_t fingerprint_len,
    const uint32_t *child_path, size_t child_path_len)
```

src/psbt.c
```diff
@@ -1436,6 +1460,36 @@ static int psbt_input_from_tx_input(struct wally_psbt *psbt,
...
+
+    return wally_psbt_input_taproot_keypath_add(&psbt->inputs[index],
+                                                pub_key,
+                                                pub_key_len,
+                                                tapleaf_hashes,
+                                                tapleaf_hashes_len,
+                                                fingerprint,
+                                                fingerprint_len,
+                                                child_path,
+                                                child_path_len);
```
jgriffiths, line 1490: for calls you can wrap/pair up the call args, e.g.:

```
    return wally_psbt_input_taproot_keypath_add(&psbt->inputs[index],
                                                pub_key, pub_key_len,
                                                tapleaf_hashes, tapleaf_hashes_len,
                                                fingerprint, fingerprint_len,
                                                child_path, child_path_len);
```

src/test/test_psbt.py
```diff
@@ -308,6 +309,46 @@ def test_invalid_args(self):
...
+            (psbt, index, 0x01, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len), # invalid flags
+            (psbt, index, flags, pubkey, pubkey_len + 1, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len), # wrong pubkey length
+            (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len - 1, fpr, fpr_len, child_path, child_path_len), # wrong tapleaf_hashes_len
+            (psbt, index, flags, pubkey, pubkey_len, None, merkle_len, fpr, fpr_len, child_path, child_path_len), # merkle length should be 0
+            (psbt, index, flags, None, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len), # no pubkey given
+            (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len - 1, child_path, child_path_len), # wrong fpr length
+            (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len - 1), # wrong child path length
+        ]
+
+        for arg in invalid_args:
```
jgriffiths, line 345: use `args` not `arg`

src/test/test_psbt.py
```diff
@@ -308,6 +309,46 @@ def test_invalid_args(self):
...
+        child_path = ctypes.c_uint32(0)
+        child_path = ctypes.pointer(child_path)
+        child_path_len = 1
+        index = 0
+        flags = 0
+
+        valid_arg = (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len)
+        invalid_args = [
+            (None, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len), # NULL psbt
+            (psbt, 1, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len), # invalid index
```
jgriffiths, line 335: The fact that this test fails while the index check is incorrect suggests something else is wrong in the args. If you set index to 0 without changing anything else, this test should fail. Does that happen? if not, you need to correct whichever other argument is causing it to fail throughout the test cases.

src/test/test_psbt.py
```diff
@@ -308,6 +309,46 @@ def test_invalid_args(self):
...
+            (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len - 1, child_path, child_path_len), # wrong fpr length
+            (psbt, index, flags, pubkey, pubkey_len, merkle, merkle_len, fpr, fpr_len, child_path, child_path_len - 1), # wrong child path length
+        ]
+
+        for arg in invalid_args:
+            self.assertEqual(WALLY_EINVAL, wally_psbt_add_input_taproot_keypath(*arg))
+            self.assertEqual(WALLY_EINVAL, wally_psbt_add_output_taproot_keypath(*arg))
+
+        self.assertEqual(WALLY_OK, wally_psbt_add_input_taproot_keypath(*valid_arg))
+        self.assertEqual(WALLY_OK, wally_psbt_add_output_taproot_keypath(*valid_arg))
```
jgriffiths, line 350: `valid_args`

src/test/test_psbt.py
instagibbs: I think it's due to out of bounds access in wally_psbt_add_input_taproot_keypath triggering a further WALLY_EINVAL

instagibbs:
rebased on https://github.com/ElementsProject/libwally-core/pull/394

jgriffiths:
Merging under https://github.com/ElementsProject/libwally-core/pull/394 with some minor changes (docs, Java wrapper updates). Thanks @instagibbs

src/map.c
```diff
@@ -620,6 +620,18 @@ int wally_keypath_xonly_public_key_verify(const unsigned char *key, size_t key_l
...
 }
 
+int wally_merkle_path_xonly_public_key_verify(const unsigned char *key, size_t key_len,
+                                              const unsigned char *val, size_t val_len)
+{
+    struct ext_key extkey;
+
+    if (key_len != EC_XONLY_PUBLIC_KEY_LEN ||
+        keypath_key_verify(key, key_len, &extkey) != WALLY_OK ||
+        extkey.version || !val || !val_len || val_len % SHA256_LEN != 0)
```
instagibbs, line 630: hmmm, with `!val_len`,  this means you can never call e.g., `wally_psbt_input_taproot_keypath_add` with a merkle length of 0.

Worked around this by adding to destination map `taproot_leaf_paths` directly via `wally_map_keypath_add`

jgriffiths: As discussed offline, the BIP says we have to support the root here (no hashes, i.e. 0-length), so I'll merge a fix for this to master.

PR #391, opened 2023-05-04
Misc fixes/updates

jgriffiths:
Minor fixes and stack usage reduction for descriptors, initial bip85 support.


JamieDriver:
Eyeballed code.

Early dev/testing w/ miniscript in Jade suggests these changes halve the stack usage compared to master.
eg. `wsh(or_d(pk(A),and_v(v:multi(2,B,C),older(4320))))` uses ~8k to generate and address, rather than ~16kb.

Agree this now works as hoped/expected: `or_d(thresh(1,pk(A)),and_v(v:thresh(1,pk(B)),older(30)))`.

Agree changes/improvements to bip85 impl - many thanks for sorting autogenerated wrappers & fixing test issues.
Locally flipped Jade to use wally bip85 impl & all good, jade tests pass unchanged.

On that basis, ack e9e3dc18

PR #392, opened 2023-05-08
descriptor updates

jgriffiths:
Allow limiting and querying the maximum nesting depth of a descriptor. Make the substitution variables compatible with other map usage.

cc: @Sjors @k-matsuzawa please note the ABI changes.


JamieDriver:
Eyeballed code.
Tested map vars change - agree all good.
Tested getting script via python wrapper - agree fixed.
Tested 'descriptor_get_depth()'  and the parsing depth limit - both lgtm.
On this basis, ack cf9bffa3

PR #394, opened 2023-05-11
Asset coin selection + misc fixes

jgriffiths:
Some misc fixes including fixing the wheel CI build. Update secp-zkp to the latest master.

Adds `wally_coinselect_assets` for performing optimized coin selection for assets (aggressive solution space reduction, checks >600M permutations/sec on Ryzen 7 5700U).

Includes a squashed/slightly updated https://github.com/ElementsProject/libwally-core/pull/390 for merging.


src/coins.c
```diff
@@ -0,0 +1,167 @@
...
+ *   Finding such a solution other than a trivial one in computationally
+ *   feasible time is highly improbable anyway.
+ * - We assume the outputs will be blinded.
+ * - The same asset is usually/mostly received to the same wallet/on the same
+ *   address type and so we can ignore effective value in favour of absolute
+ *   value.
+ * - Dust is not meaningful for assets, since we cannot know their value
+ *   relative to L-BTC. We nevertheless attempt to use only the largest UTXOs.
+ * - Privacy is somewhat protected by blinding of the asset and value, so we
+ *   do not care about spending the same addresses together.
```
LeoComandini, line 26: If there are outputs with the same scriptpubkey, not spending them together allows to link transactions, which is a privacy issue.

However in elements if we spend all utxos (of any asset) which have the same scriptpubkey, an attacker might cheaply send assets without value to the target, which will have to waste fees for the extra tx inputs and (perhaps) outputs.

OTOH if the wallet has only assets which have some value, then spending all utxos with the same scriptpubkey is better from a privacy perspective.

We might not want to do it for complexity and costs reasons though.

src/swig_python/python_extra.py_in
```diff
@@ -16,13 +16,19 @@ def _wrap_bin(fn, length, resize=False):
...
     def wrapped(*args):
         n = length(*args) if callable(length) else length
         out = [None] * n
         ret = fn(*list(args)+[out])
-        assert ret == len(out) # Input list must be of the correct size
+        if resize:
+            # Truncate as per _wrap_bin
+            assert ret <= n
+            return out[0:ret] if ret != n else out
+        if ret is not None:
```
LeoComandini, line 29: Why do we need this?

LeoComandini:
utACK 58669b8a38eeffd22b2e58638b0bfbb0f9d5cb06, with minor comment, core review.

Perhaps we want someone else to have a look at the last commit, I'm not particularly confident of my review.
Or maybe we are fine with testing it via GDK.

src/swig_python/python_extra.py_in
jgriffiths: There was a duplicated assert which we didn't need and I've just removed, good spotting.

The `None` check is needed for any future wrapper that returns a fixed size `uint32_t` array, but we want to wrap with a potential callable to determine its output size from its inputs.

jgriffiths:
> Perhaps we want someone else to have a look at the last commit, I'm not particularly confident of my review. Or maybe we are fine with testing it via GDK.

I had the algorithm and its assumptions pre-reviewed by @instagibbs prior to pushing this PR. A gdk change to switch to this call for asset UTXO selection is WIP and will be tested there.

Note that I've done a fair amount of testing much of which isn't really amenable to being tested in wally, given many outlier tests (e.g. huge UTXO sets with very large io_ratios) can take a very long time to run.

src/coins.c
jgriffiths: > spend all utxos (of any asset) 

This will never be a good idea AFAICS. Its one thing to scoop up related inputs, but a very different proposition to create new outputs. I think the best cause of action for users receiving unwanted/spam assets is to blacklist those UTXOs or ignore them (since they shouldn't be able to be spent accidentally).

> spending all utxos with the same scriptpubkey is better from a privacy perspective

I think this case is rare enough that it can be handled outside of the the exposed wally call if the caller wishes to, e.g. if the caller determines that some number of utxos share a spk:

- First try a normal selection. If a solution is found that includes either all or none of the same-spk inputs, then use it.
- Otherwise, combine the same-spk inputs together into a single value, and perform a selection with io_ratio increased.

I'm expecting there may be alternate selection algorithms in the future, some of which could take extra data like the spk or a tag to mark inputs that should be spent together. However I don't want to complicate the standard logic for this scenario.

src/test/test_coinselection.py
```diff
@@ -0,0 +1,31 @@
...
+            (values, 0,          target, attempts,   ratio, out,  out_len),   # Empty values
+            (values, values_len, 0,      attempts,   ratio, out,  out_len),   # Zero target
+            (values, values_len, target, values_len, ratio, out,  out_len),   # Too few attempts
+            (values, values_len, target, attempts,   0,     out,  out_len),   # Zero ratio
+            (values, values_len, target, attempts,   ratio, None, out_len),   # Null output
+            (values, values_len, target, attempts,   ratio, out,  out_len-1), # Output too small
+        ]
+        for args in bad_args:
+            ret = wally_coinselect_assets(*args)
+            self.assertEqual(ret, (WALLY_EINVAL, 0))
```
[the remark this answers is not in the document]
jgriffiths: Good idea, added.

PR #402, opened 2023-07-07
Descriptor updates

jgriffiths:
Add support for generation of wrappers with value expressions. Verify that adding this support does not expose a recent vulnerability affecting Ledger in this case. Also add support for `raw()` to generate empty scripts.

Note that the test cases here expose what appear to be issues in either the spec or core implementation. Until such time as the spec is confirmed incorrect, the behaviour of the wally impl is expected to/tested to follow it.

cc: @k-matsuzawa @JamieDriver


JamieDriver:
Did a bit of testing with 'a:1' expression and seems good.  
eyeballed code, can't see anything glaring, although not very familiar with this area.
ack [5220c20](https://github.com/ElementsProject/libwally-core/pull/402/commits/5220c2074a53f42c176d396a55db77eef89c31a6)

PR #411, opened 2023-08-30
configure.ac: fix Bashisms

whitslack:
The POSIX Shell `test` built-in does not define a `==` operator. It works if `/bin/sh` happens to be Bash, but that is non-standard.

@jgriffiths: Since you're planning a 1.0.0 release, now would be a good time to fix this so I can finally discontinue patching it downstream.


jgriffiths:
Thanks!

PR #414, opened 2023-08-31
support `--with-system-secp256k1`

whitslack:
Libwally-core, in its stock configuration, compiles and statically links against a private copy of libsecp256k1_zkp. However, Gentoo unbundles libsecp256k1_zkp and instead links libwally-core against a system-wide shared libsecp256k1_zkp with headers in `/usr/include/secp256k1_zkp` and a `libsecp256k1_zkp.pc` that specifies the relevant `CFLAGS` and `LIBS`.

Implement a `--with-system-secp256k1` configure option to allow compiling and linking against a system-installed libsecp256k1. Pass the user-specified package name (or `libsecp256k1` or `libsecp256k1_zkp` by default, depending on `--enable-standard-secp`) to `PKG_CHECK_MODULES` to find the `CFLAGS` and `LIBS` of a system-installed libsecp256k1. Call `AC_CHECK_FUNCS` to assert that the required modules are present in the system-installed libsecp256k1.

**If the user does not specify `--with-system-secp256k1` (or specifies `--without-system-secp256k1`), then the pre-existing behavior is preserved: namely, the build will use the bundled copy of libsecp256k1_zkp.**

Adjust several `#include` directives to reference secp256k1 headers in the configured header search path rather than explicitly specifying paths to the private copies. In the case of using the bundled libsecp256k1_zkp, `src/secp256k1/include` is added to the header search path; otherwise, the necessary `CFLAGS`, if any, are taken from pkg-config.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -259,29 +259,49 @@ fi
 #
 # libsecp256k1
 #
-libsecp256k1_CFLAGS='-I$(top_srcdir)/src/secp256k1/include'
-AC_SUBST([libsecp256k1_CFLAGS])
+AC_ARG_WITH([system-secp256k1],
+    [AS_HELP_STRING([[--with-system-secp256k1[=PKG]]],
+        [build using system-installed libsecp256k1 instead of bundled, passing PKG (default: libsecp256k1 or libsecp256k1_zkp, depending on --enable-standard-secp) to pkg-config (default: no)])],
+    [AS_IF([test "x$withval" = xyes],
+        [AM_COND_IF([BUILD_STANDARD_SECP], [with_system_secp256k1=libsecp256k1], [with_system_secp256k1=libsecp256k1_zkp])])],
+    [with_system_secp256k1=no])
 
-# FIXME: This is needed to force libtool to use all object files from secp.
-#        We can only build secp properly by recursively invoking
-#        configure/make, and can't include it as a noinst_ library. Libtool
-#        assumes that such libraries will be installed along with our library
-#        target and so won't force all object files in the library to be
-#        included in ours - despite the fact that we are making a shared
-#        library and linking to a static one. This is broken and we work
-#        around it by hacking the secp objects directly into the library
-#        via the _LDADD variable for wallycore.
-#        We previously achieved this by adding the libsecp256k1.a archive,
-#        but changes to libtool and apples linkers mean that
-#        archives-within-archives no longer work.
-#        Because automake tries to police its users very strictly and fails
-#        hard when flags are passed in this way, we have to substitute the
-#        flags here.
-#        Because libtool both intercepts -Wl and arbitrarily re-orders its
-#        command line inputs, we have to concoct a single expression to
-#        enforce linking that cannot be split, hence the below expression.
-LIBADD_SECP256K1="-Wl,secp256k1/src/libsecp256k1_la-secp256k1.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult_gen.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult.${OBJEXT}"
-AC_SUBST([LIBADD_SECP256K1])
+AM_CONDITIONAL([LINK_SYSTEM_SECP256K1], [test "x$with_system_secp256k1" != xno])
+AM_COND_IF([LINK_SYSTEM_SECP256K1], [
+    dnl Use the secp installed system-wide (after checking it for suitability)
+    saved_LIBS=$LIBS
+    m4_ifdef([PKG_CHECK_MODULES],
+        [PKG_CHECK_MODULES([libsecp256k1], [$with_system_secp256k1])],
+        [AC_MSG_ERROR([You need to install pkg-config to use --with-system-secp256k1.])])
+    LIBS="$libsecp256k1_LIBS $LIBS"
+    missing_modules=
+    AC_DEFUN([CHECK_MODULE], [
+        AC_CHECK_FUNCS([$2], [], [missing_modules="${missing_modules} $1"])
+    ])
+    CHECK_MODULE([ecdh], [secp256k1_ecdh])
+    CHECK_MODULE([extrakeys], [secp256k1_xonly_pubkey_parse])
+    CHECK_MODULE([recovery], [secp256k1_ecdsa_recover])
+    CHECK_MODULE([schnorrsig], [secp256k1_schnorrsig_verify])
+    AM_COND_IF([BUILD_STANDARD_SECP], [], [
+        CHECK_MODULE([ecdsa-s2c], [secp256k1_ecdsa_s2c_sign])
+    ])
+    AM_COND_IF([BUILD_ELEMENTS], [
+        CHECK_MODULE([generator], [secp256k1_generator_parse])
+        CHECK_MODULE([rangeproof], [secp256k1_rangeproof_verify])
+        CHECK_MODULE([surjectionproof], [secp256k1_surjectionproof_initialize])
+        CHECK_MODULE([whitelist], [secp256k1_whitelist_sign])
+    ])
+    AS_IF([test -n "${missing_modules}"], [
+        AC_MSG_ERROR([system-installed $with_system_secp256k1 does not support these required modules:${missing_modules}])
+    ])
+    LIBS=$saved_LIBS
+], [
+    dnl Use the secp in-tree submodule
+    libsecp256k1_CFLAGS='-I$(top_srcdir)/src/secp256k1/include'
+    libsecp256k1_LIBS='$(top_srcdir)/src/secp256k1/libsecp256k1.la'
+])
+AC_SUBST([libsecp256k1_CFLAGS])
+AC_SUBST([libsecp256k1_LIBS])
 
 #
 # Python facilities
@@ -408,6 +428,8 @@ export AR_FLAGS
 export LD
 export LDFLAGS
 
-AX_SUBDIRS_CONFIGURE([src/secp256k1], [[--disable-shared], [--enable-static], [--with-pic], [--enable-experimental], [--enable-module-ecdh], [--enable-module-recovery], [--enable-module-ecdsa-s2c], [--enable-module-rangeproof], [--enable-module-surjectionproof], [--enable-module-whitelist], [--enable-module-generator], [--enable-module-extrakeys], [--enable-module-schnorrsig], [$secp256k1_test_opt], [--enable-exhaustive-tests=no], [--enable-benchmark=no], [--disable-dependency-tracking], [$secp_asm]])
+AM_COND_IF([LINK_SYSTEM_SECP256K1], [], [
+	AX_SUBDIRS_CONFIGURE([src/secp256k1], [[--disable-shared], [--enable-static], [--with-pic], [--enable-experimental], [--enable-module-ecdh], [--enable-module-recovery], [--enable-module-ecdsa-s2c], [--enable-module-rangeproof], [--enable-module-surjectionproof], [--enable-module-whitelist], [--enable-module-generator], [--enable-module-extrakeys], [--enable-module-schnorrsig], [$secp256k1_test_opt], [--enable-exhaustive-tests=no], [--enable-benchmark=no], [--disable-dependency-tracking], [$secp_asm]])
+])
 
 AC_OUTPUT
--- a/configure.ac
+++ b/configure.ac
@@ -407,8 +407,7 @@ export ARFLAGS
 export AR_FLAGS
 export LD
 export LDFLAGS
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -258,6 +258,50 @@ fi
...
+    [AS_IF([test "x$withval" = xyes],
+        [AM_COND_IF([BUILD_STANDARD_SECP], [with_system_secp256k1=libsecp256k1], [with_system_secp256k1=libsecp256k1_zkp])])],
+    [with_system_secp256k1=no])
+AM_CONDITIONAL([LINK_SYSTEM_SECP256K1], [test "x$with_system_secp256k1" != xno])
+AM_COND_IF([LINK_SYSTEM_SECP256K1], [
+    saved_LIBS=$LIBS
+    PKG_CHECK_MODULES([libsecp256k1], [$with_system_secp256k1])
+    LIBS="$libsecp256k1_LIBS $LIBS"
+    AM_COND_IF([BUILD_STANDARD_SECP], [
+        AC_CHECK_FUNCS([
```
jgriffiths, line 273: I think this might be cleaner if its broken into:

- check for the standard modules (for all configs)
- if non-standard secp, check for ecdsa-s2c
- if elements, check for generator, rangeproof, surjectionproof, and whitelist

This will be easier to update going forward when new modules are required.

configure.ac
```diff
@@ -276,8 +320,12 @@ fi
 #        Because libtool both intercepts -Wl and arbitrarily re-orders its
 #        command line inputs, we have to concoct a single expression to
 #        enforce linking that cannot be split, hence the below expression.
-LIBADD_SECP256K1="-Wl,secp256k1/src/libsecp256k1_la-secp256k1.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult_gen.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult.${OBJEXT}"
-AC_SUBST([LIBADD_SECP256K1])
+    libsecp256k1_LIBS="-Wl,secp256k1/src/libsecp256k1_la-secp256k1.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult_gen.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult.${OBJEXT}"
+    ac_configure_args="${ac_configure_args} --disable-shared --with-pic --with-bignum=no --enable-experimental --enable-module-ecdh --enable-module-recovery --enable-module-ecdsa-s2c --enable-module-rangeproof --enable-module-surjectionproof --enable-module-whitelist --enable-module-generator --enable-module-extrakeys --enable-module-schnorrsig ${secp256k1_test_opt} --enable-openssl-tests=no --enable-exhaustive-tests=no --enable-benchmark=no --disable-dependency-tracking ${secp_asm}"
```
jgriffiths, line 324: unrelated to this change, but `--enable-openssl-tests` is no longer supported in secp and should be removed (ideally in its own commit).

Setting these args before we initialize secp_asm/secp256k1_test_opt and export the various variables seems wrong, what am I missing?

configure.ac
whitslack: The only reason I didn't do it that way was to improve the error messages. The first error encountered will abort the configure, yet the message will not report all needed modules. (This is true presently, but your suggestion makes it even worse.) But I agree in principle. I will pursue a cleaner way of checking for all needed modules at once and reporting all in a single error message by first building lists of needed modules, depending on enabled options, before performing the check.

configure.ac
whitslack: > `--enable-openssl-tests` is no longer supported in secp and should be removed (ideally in its own commit).

Indeed. Really in its own PR, as it's off-topic for this PR.

> Setting these args before we initialize secp_asm/secp256k1_test_opt and export the various variables seems wrong

You're right. This slipped past me. I'll move the `ac_configure_args` and `AC_CONFIG_SUBDIRS` to the bottom again, guarded by a conditional on `!LINK_SYSTEM_SECP256K1`.

Incidentally, abusing `ac_configure_args` in this way is problematic if the `Makefile` determines that it's out of date and needs to call `autoreconf` (et al.) to regenerate itself, as then it calls back to `configure` to reconfigure libwally-core, but at that point all these extraneous configure options, intended for libsecp256k1, are appended to the libwally-core configure line. Especially troublesome is tho `--disable-shared`, which causes the libwally-core build to produce a static library even when libwally-core was originally configured with `--enabled-shared`.

Have you considered using [`AX_SUBDIRS_CONFIGURE`](https://www.gnu.org/software/autoconf-archive/ax_subdirs_configure.html), which is designed to provide for exactly this use case of configuring an third-party sub-project with different configure options than the enclosing project? I almost went that route myself but decided to try to preserve what you already had in place so as to maximize the chances of getting my PR merged.

configure.ac
```diff
@@ -259,6 +259,40 @@ fi
...
+    ])
+    AM_COND_IF([BUILD_ELEMENTS], [
+        CHECK_MODULE([generator], [secp256k1_generator_parse])
+        CHECK_MODULE([rangeproof], [secp256k1_rangeproof_verify])
+        CHECK_MODULE([surjectionproof], [secp256k1_surjectionproof_initialize])
+        CHECK_MODULE([whitelist], [secp256k1_whitelist_sign])
+    ])
+    AS_IF([test -n "${missing_modules}"], [
+        AC_MSG_ERROR([system-installed $with_system_secp256k1 does not support these required modules:${missing_modules}])
+    ])
```
whitslack, line 292: @jgriffiths: Do you prefer this method of checking for required libsecp256k1 modules?

configure.ac
jgriffiths: Thanks. As a maintainer my priority is to ensure its as easy as possible to see where I need to add new module dependencies (and have that be a single place). If that comes at the cost of bespoke builds requiring multiple configure runs, its a price I'm prepared to pay :)

configure.ac
jgriffiths: AX_SUBDIRS_CONFIGURE looks like a cleaner way to handle secp configuring, I wasn't aware of it. I'm amenable to testing it if you'd like to go in that direction.

configure.ac
jgriffiths: Yes that works for me with my maintainer hat on, thanks.

jgriffiths:
@whitslack you'll need to rebase, I'm keeping the docs updates and version bump floating as the top commits.

setup.py
```diff
@@ -51,13 +51,15 @@
 
     abs_path = os.path.dirname(os.path.abspath(__file__)) + '/'
 
-    def call(args):
-        subprocess.check_call(args, cwd=abs_path, env=configure_env)
+    if not os.path.isfile(abs_path + 'src/.libs/libwallycore' + ('.so' if build_shared else '.a')):
```
whitslack, line 54: `.so` isn't the suffix of shared libraries on macOS.

@jgriffiths: Do you have any suggestion of how to check here whether the shared library has already been built?

In my Gentoo ebuild, I've been having to use `sed` to delete this entire `if not is_windows:` block because we build the Python module (conditionally, depending on the `python` `USE` flag) after the C library has already been built, but `setup.py` always tries to rebuild it. Worse, Gentoo's Python build machinery makes multiple calls to `setup.py` since building and installing are separate ebuild phases, yet this `setup.py` does the build unconditionally every time it's executed. So that's my motivation for making the build steps here conditional depending on whether the needed library type already exists.

jgriffiths: It should be possible to determine what args setup.py was called with I think, can you give me the calls that are made (and the arguments) by the Gentoo machinery?

whitslack: > can you give me the calls that are made (and the arguments) by the Gentoo machinery?

Hmm, actually no, I can't. It's complicated by the fact that `setup.py` is considered deprecated, so Gentoo now uses [gpep517](https://pypi.org/project/gpep517/) in legacy setuptools mode to drive the execution of the code in `setup.py` for projects that have not yet graduated to a PEP517-compliant build system. The extra layer of abstraction makes it all very unclear to me. (I'm most definitely not a Python guy.)

whitslack: If you don't like the `isfile` conditional here, I'm fine to drop it, and my ebuild will continue to sed out the code block containing these build commands. Note, however, that that would make it awkward for anyone else who is trying to achieve the "full dynamic link" outcome, as the unmodified `setup.py` will undo the effect of `--with-system-secp256k1` and rebuild the library with embedded secp256k1 code.

jgriffiths: I'm focused on getting the other PR in first, but in general I'm less worried about building the python wheel being more complicated in this case (not to say I'm not willing to add support).

whitslack: In general, what is the logic behind having the Python module be responsible for building the C library? That's an unexpected inversion of responsibility relative to historical norms, wherein the native library is built standalone, and then wrappers for other languages are built to link with the native library.

jgriffiths: We've had many problems in the past not using the Python build machinery, as Python doesn't care about historic norms. Python expects the c code for c-extensions to be built by setup.py, and although you can compile just the wrapping sources and link with a DSO as you will on Gentoo, that only works for distributions and can't work for PyPI installs as (a) pip installing wallycore would import an unusable module unless you had a compatible wally shared lib installed and (b) it circumvents the security of using a requirements.txt version hash because you can't be sure which libwallycore.so is going to be loaded.

whitslack: > pip installing wallycore would import an unusable module unless you had a compatible wally shared lib installed

That's exactly what I would expect, to be frank. Trying to use a wrapper library without also having installed the underlying library that it wraps should be expected to fail.

> it circumvents the security of using a requirements.txt version hash because you can't be sure which libwallycore.so is going to be loaded.

This amuses me because I see it just the opposite: pinning a version of libwally-core in the Python wrapper means that the Python wrapper *cannot* benefit from security fixes in the underlying libwally-core unless the Python wrapper also gets updated. A user may upgrade libwally-core to resolve a security issue and expect that all code which uses libwally-core will be fixed. That's certainly the assumption on Gentoo, which is why static linking is so insidious and harmful to security. I guess it's like tabs versus spaces: we'll never agree. (Incidentally, your code base is horrible — I keep having to hit SPACE SPACE SPACE SPACE in my editors since you don't use sane tabulation. 😁)

So what do you want to do about the matter at hand? Like I said, I'm okay to rip out the conditional and keep patching it downstream, which sounds like the right thing to do since linking a Python wrapper module dynamically to the library that it wraps sounds like practice unsanctioned by the Python gods.

jgriffiths:
@whitslack I've merged the core change for `--with-system-secp256k1` and removed the commit comment about the missing secp dependency being irrelevant as its clearly a libtool bug and users may force static linking in which case the missing dependency is relevant. I've documented instead that users should manually link with secp when statically linking.

I think this just leaves the python/setup.py changes outstanding. If you can close these two outstanding PRs, rebase and put the Python changes into a single PR then I'll look at how to merge them. Note I won't be changing the amalgamation compile by default, but it should be possible to allow you to build without it via env var or similar.

whitslack:
> Note I won't be changing the amalgamation compile by default, but it should be possible to allow you to build without it via env var or similar.

@jgriffiths: Can you clarify? It doesn't make sense to include *both* the amalgamated sources *and* `-lwallycore`.

> `PKG_CHECK_MODULES` breaks the configure script if `pkg-config` is not installed

Who the heck is building a software package from source without having `pkg-config` installed? Lolwhut?! That's almost as essential as `make`. But hmm, I'll see if I can hack around it. 🤔

whitslack:
> I've documented instead that users should manually link with secp when statically linking.

> When building as a static library, the user will need to link libsecp256k1.a in addition to libwallycore.a.

This is a little misleading. Preferably users shouldn't be specifying any `.a` files when linking against a libtool-built library; they should specify `libwallycore.la`, which will automatically handle adding both the `libwallycore.a` and the required `libsecp256k1.a` to the link. It's only for non-libtool build systems that one would need to manually specify dependency libraries.

whitslack:
> `PKG_CHECK_MODULES` breaks the configure script if `pkg-config` is not installed

25e1658dfb82fa58aa4474c5f90566a046d66a16 should address this.

jgriffiths:
> they should specify libwallycore.la, which will automatically handle adding both the libwallycore.a and the required libsecp256k1.a to the link.

As above, this is broken with libtool if they build dynamic and static, but want to link statically. But also, not everyone uses libtool to link their app, and wally should not force that upon them.

whitslack:
> As above, this is broken with libtool if they build dynamic and static, but want to link statically.

True, but isn't that a moot point? If both shared and static libraries are found, `-lwallycore` will prefer the shared one.

Anyway, I have no real point. If someone uses libtool, then they ought to know that they're supposed to link against only `libwallycore.la` and let libtool handle the dependency, so it doesn't need to be spelled out to them.

jgriffiths:
Hi @whitslack sorry for the delay. I've cherry-picked and merged most of the 1.0.0 changes, with https://github.com/ElementsProject/libwally-core/pull/422 scheduled to be merged next.

Can you rebase this on https://github.com/ElementsProject/libwally-core/pull/422 and include your proposed fixup commit https://github.com/ElementsProject/libwally-core/commit/25e1658dfb82fa58aa4474c5f90566a046d66a16 ? the CI will then prove whether the build is working without pkgconfig installed. Once this commit is merged all that remains is the python linkage PR which I'll look at once this change is in, thanks.

whitslack:
> Can you rebase this on #422 and include your proposed fixup commit [25e1658](https://github.com/ElementsProject/libwally-core/commit/25e1658dfb82fa58aa4474c5f90566a046d66a16) ?

Done.

jgriffiths:
Merged under https://github.com/ElementsProject/libwally-core/pull/423, closing.

PR #416, opened 2023-09-01
drop `--enable-openssl-tests` from libsecp256k1 config

whitslack:
Quoting jgriffiths:
> `--enable-openssl-tests` is no longer supported in secp and should be removed

Suggested-by: @jgriffiths
See: https://github.com/ElementsProject/libwally-core/pull/414#discussion_r1312347724


jgriffiths:
Merged, thanks!

PR #417, opened 2023-09-01
clean up linkage of libsecp256k1 into libwallycore and Python extension

whitslack:
This PR has two remaining objectives:

* Eliminate the FIXME in `configure.ac` relating to embedding the symbols from `libsecp256k1.a` when building `libwallycore.so`.
* Adding an option to `setup.py` to link the Python native extension dynamically with a libwally-core DSO.


whitslack:
@jgriffiths: I noticed one regression when switching from `AC_CONFIG_SUBDIRS` to `AX_SUBDIRS_CONFIGURE`: `autoreconf` no longer recurses into `src/secp256k1`, meaning the Autotools infrastructure needs to be manually (re)generated in `src/secp256k1`. I am not sure if there will be an impact to your automated testing, continuous integration, and/or release engineering processes due to this regression.

whitslack:
> I am not sure if there will be an impact to your automated testing, continuous integration, and/or release engineering processes due to this regression.

Indeed, there was a break in that `setup.py` was relying on the top-level `tools/autogen.sh` to recursively run `autoreconf` in `src/secp256k1`. That no longer happens, so I added an explicit call to `src/secp256k1/autogen.sh` in `setup.py`.

configure.ac
```diff
@@ -259,26 +259,10 @@ fi
...
-#        flags here.
-#        Because libtool both intercepts -Wl and arbitrarily re-orders its
-#        command line inputs, we have to concoct a single expression to
-#        enforce linking that cannot be split, hence the below expression.
-LIBADD_SECP256K1="-Wl,secp256k1/src/libsecp256k1_la-secp256k1.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult_gen.${OBJEXT},secp256k1/src/libsecp256k1_precomputed_la-precomputed_ecmult.${OBJEXT}"
-AC_SUBST([LIBADD_SECP256K1])
+libsecp256k1_CFLAGS='-I$(top_srcdir)/src/secp256k1/include'
+libsecp256k1_LIBS='$(top_srcdir)/src/secp256k1/libsecp256k1.la'
+AC_SUBST([libsecp256k1_CFLAGS])
+AC_SUBST([libsecp256k1_LIBS])
```
whitslack, line 265: This fix might bear some discussion. The assertion “Libtool […] won't force all object files in the library to be included in ours - despite the fact that we are making a shared library and linking to a static one. This is broken” appears to be incorrect. When I link `libwallycore.la` with libtool in shared mode (i.e., configured with `--enable-shared`) to produce `libwallycore.so`, passing in the `libsecp256k1.la` that was just built in static mode, libtool _does_ copy in all the symbols from the static library and does export the symbols with public visibility in the resulting shared library's dynamic symbol table. In fact I verified using `diff` that the set of `secp256k1_*` dynamic symbols exported from the resulting `libwallycore.so` exactly matches the set of `secp256k1_*` dynamic symbols exported from a `libsecp256k1.so` that results from the same libsecp256k1 sources rebuilt in shared mode, so I don't know how the confusion originated. Maybe there was once a defect in libtool at some time in the past, but if so, it appears to have been fixed.

jgriffiths: I believe this does not affect all platforms, IIRC Mac and BSD were affected.

whitslack: I don't have any way to test BSD, but I could conceivably test macOS if it's not convenient for you to do it.

whitslack: Okay, so I tested on macOS 10.13, and libtool does the same thing as it does on Linux:

* If configured to build a _static_ libwally-core library, libtool **does not** embed the secp256k1 symbols into `libwallycore.a`, but it **does** list the path of the just-built `libsecp256k1.la` on the `dependency_libs` line in `libwallycore.la`.
* If configured to build a _shared_ libwally-core library, libtool **does** embed the secp256k1 symbols into `libwallycore.1.dylib`, and it **does not** list anything on the `dependency_libs` line in `libwallycore.la`.

So I don't see any problem.

I can't test any BSDs, but I would expect them to have the same behavior as macOS.

jgriffiths:
> That no longer happens, so I added an explicit call to src/secp256k1/autogen.sh in setup.py.

I think this explicit call should be added to `tools/autogen.sh` instead of in setup.py. This will then not affect the other users of the library who build wally according to the README instructions (in particular the Green wallets and c-lightning).

tools/autogen.sh
```diff
@@ -8,3 +8,7 @@ if uname | grep "Darwin" >/dev/null 2>&1; then
         done
     done
 fi
+
+if [ -x src/secp256k1/autogen.sh ] ; then
+    cd src/secp256k1 && ./autogen.sh
+fi
```
whitslack, line 14: @jgriffiths: Better?

jgriffiths:
cherry-picked e14ada812e8aa720fe29740563a86d788e34a291 to reduce rebase pain.

configure.ac
jgriffiths: This workaround was put in place in 2016, so its certainly possible the underlying bug(s) were fixed. I think we have an openbsd box in our internal CI somewhere, I'll need to check the build on that.

tools/autogen.sh
jgriffiths: Perfect, thanks

jgriffiths:
@whitslack can you rename `WALLY_BUILD_SHARED` to `WALLY_ABI_PY_WHEEL_SHARED`?

whitslack:
> can you rename `WALLY_BUILD_SHARED` to `WALLY_ABI_PY_WHEEL_SHARED`?

@jgriffiths: Hmm, yes, but is that a clear name for what the variable controls? To me, that name suggests that the variable controls whether the _wheel_ is to be shared, which doesn't entirely make sense. What we're actually talking about here is whether the Python native extension library (which is always a shared library included in the wheel) will link dynamically to libwally-core (or else will embed the libwally-core and libsecp256k1 code wholly within the Python native extension library).

jgriffiths:
Naming things, always the hardest part of coding lol.

How about `WALLY_ABI_PY_WHEEL_USE_DSO`? indicating that the wheel is using dynamic shared objects?

whitslack:
> `WALLY_ABI_PY_WHEEL_USE_DSO`

Okay, sure. I have no strong objections to that name. It's a little verbose, but it's okay.

jgriffiths:
In the first commit you need to include the m4 file defining `AX_SUBDIRS_CONFIGURE` from the autoconf archive, I think you've forgotten to check it in?

whitslack:
> In the first commit you need to include the m4 file defining `AX_SUBDIRS_CONFIGURE` from the autoconf archive, I think you've forgotten to check it in?

It's part of Autoconf Archive, which I think you're already using. `autoreconf` takes care of finding the macro definition beneath `/usr{,/local}/share/aclocal`.

whitslack:
> we check in the macros we need from the archive

Do you monitor the Archive for bug fixes and feature enhancements in the macros you've copied from it, and do you promptly make new releases of libwally-core whenever the macros you're using have received such a bug fix or enhancement? This is the static linking versus dynamic linking debate all over again. 😅 Maybe one of your users has an unusual setup that causes an Autoconf Archive macro to misbehave. Upon researching it, they discover that the misbehavior has already been corrected in the Archive, but because you haven't been keeping up with fixes, your project is broken for them, whereas everything would have been fine if you hadn't made your own private copy that inevitably becomes outdated. (I'm sure this comes across as though I'm very perturbed, but I am not. It's a difference of philosophy as old as the concept of linking itself. I don't have any illusions of changing your mind.) I'll make a copy of the macro definition and check it in. 🤦‍♂️

whitslack:
> I'll make a copy of the macro definition and check it in.

This is done.

configure.ac
whitslack: Another big landmine that we haven't mentioned yet is Windows. Apparently you do support that OS. I have no idea what libtool does with DLLs on such a system. I have no way to test this either, as I haven't used Windows for a couple of decades at this point. 😕

jgriffiths: autotools on windows isn't worth supporting; too many special cases, bugs and failure modes. For Python we have a build that works by including the amalgamation, downstream app users are expected to statically link given the security and versioning hell that is windows DSO support.

jgriffiths:
> I don't have any illusions of changing your mind

Note the requirements for wally, exposed functionality, how it is expected to be used and the time available for me to work on it are all determined by my day job; my personal opinions on the correct way of doing things is irrelevant and I'm not arguing from my POV at all. I'm not adverse to explaining the rationale behind these decisions as I understand them but your views (however correct they may be) don't supersede my work obligations.

I'll be removing some verbage from your commit descriptions before finalizing the build_updates branch, I will call for final review and get an ack from you before merging this branch (which may take some time given it needs further work and internal testing).

whitslack:
> I'll be removing some verbage from your commit descriptions

If you remove the paragraph beginning "At the insistence of this project's maintainer," I would respectfully request that you `git commit --amend --reset-author`. I don't want to take the blame for creating technical debt, but I am completely fine with being unattributed. In other words, I don't care about getting credit for my work, but I do care about not being blamed for doing something that I opposed. Make sense?

configure.ac
whitslack: I resurrected the `src/amalgamation/combined*.c` sources but relegated them to the `is_windows` pile only.

jgriffiths:
> I would respectfully request that you git commit --amend --reset-author.

Will do.

configure.ac
jgriffiths: Yeah I'm still pondering the changes to `setup.py`. Note the commit description for the first commit needs changing now that the secp includes have been pulled into their own commit. Also if you move the setting of `libsecp256k1_LIBS` below the existing `AC_SUBST([libsecp256k1_CFLAGS])` this diff will be a single chunk instead of 2.

jgriffiths: I'm likely to cherry pick these changes and tweak them so actually don't worry about changing this for now.

whitslack: Note, the reason for renaming `LIBADD_SECP256K1` to `libsecp256k1_LIBS` is because the latter is what gets set by `PKG_CHECK_MODULES`. Not relevant in *this* PR, but becomes relevant when `--with-system-secp256k1` is introduced.

whitslack:
Superseded by #419.

PR #418, opened 2023-09-02
compile all Java classes in one `javac` invocation

whitslack:
Starting up a JVM is not a lightweight affair. For this reason, typically all Java source files are compiled by a single invocation of `javac`. Use GNU Make's "grouped targets" feature to specify that all Java class files are produced by a single execution of the recipe. Further, avoid compiling the test classes if tests are disabled.


jgriffiths:
Nice change! merged with rebase, closing thanks.

PR #419, opened 2023-09-04
Python linkage enhancements

whitslack:
* On systems supporting Libtool (i.e., anything but Windows), skip compiling the "amalgamated" sources for the Python module and simply link in the just-built `libwallycore.a` and `libsecp256k1.a`.

* `setup.py`: support optionally building and linking to libwally-core as a dynamic shared object if `WALLY_ABI_PY_WHEEL_USE_DSO=1` is set. Ignored on Windows, where `setup.py` does not build libwally-core.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,13 @@ from setuptools import setup, Extension
 import copy, os, platform, shutil
 import distutils.sysconfig
 
-CONFIGURE_ARGS = ['--disable-shared', '--enable-static', '--with-pic',
+build_shared = os.environ.get('WALLY_ABI_PY_WHEEL_USE_DSO', '').lower() not in ['', '0', 'false', 'no', 'n', 'off']
+if build_shared:
+    CONFIGURE_ARGS = ['--enable-shared', '--disable-static']
+else:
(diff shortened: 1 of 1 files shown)


whitslack:
> It should be possible to determine what args setup.py was called with I think, can you give me the calls that are made (and the arguments) by the Gentoo machinery?

@jgriffiths: This is the command line that Gentoo's `distutils-r1.eclass` invokes to build the libwally-core Python wheel:
```
gpep517 build-wheel --backend setuptools.build_meta:__legacy__ --output-fd 3 --wheel-dir /var/tmp/portage/net-libs/libwally-core-1.0.0/work/libwally-core-release_1.0.0-python3_11/wheel
```
I'm not sure if that will be helpful to you at all. (I rather doubt it.)

whitslack:
@jgriffiths: Did this slip through the cracks, or have you decided not to support this after all?

jgriffiths:
@whitslack yes it did. The python build machinery has been updated somewhat, so this will need re-working. Please submit under a new PR if you would.

PR #428, opened 2023-10-27
bip39: expose a non-allocating bip39_get_word_by_index

### LINKED ISSUE #427: why does bip39_get_word() use the heap -- can we have a function that returns a `const char*` instead?
https://github.com/ElementsProject/libwally-core/blob/c102e7e8c7899822666e22d6cbcdaa8a1f444d38/src/bip39.c#L77

Why is the heap used here, even though the string could just be returned as is?

It would much more efficient and easier to use if we libwally exposed the `wordlist_lookup_index()` function instead, which returns the pointer to the existing string. Looping through all words right now for example requires malloc/free in each iteration.

jgriffiths:
Fixes #427

### DIFF
--- a/include/wally_bip39.h
+++ b/include/wally_bip39.h
@@ -61,6 +61,22 @@ WALLY_CORE_API int bip39_get_word(
     size_t index,
     char **output);
 
+#ifndef SWIG
+/**
+ * Get the 'index'th word from a word list.
+ *
+ * :param w: Word list to use. Pass NULL to use the default English list.
+ * :param index: The 0-based index of the word in ``w``.
+ *
+ * .. note:: Returns NULL if any argument is invalid or ``index`` is out of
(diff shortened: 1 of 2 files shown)


include/wally_bip39.h
```diff
@@ -61,6 +61,22 @@ WALLY_CORE_API int bip39_get_word(
...
     char **output);
 
+#ifndef SWIG
+/**
+ * Get the 'index'th word from a word list.
+ *
+ * :param w: Word list to use. Pass NULL to use the default English list.
+ * :param index: The 0-based index of the word in ``w``.
+ * :param output: Destination for the resulting word.
+ *
```
benma, line 71: maybe mention that this returns NULL if the index is out of bounds.

benma (approved):
LGTM, thanks a lot!

PR #431, opened 2023-11-07
mbed_tls: don't byteswap if HW acceleration is already big endian

greenaddress:
Tested on:

esp32:
- vanilla wally without mbed
- wally with mbed with acceleration off
- wally with mbed with acceleration on

esp32s3:
- vanilla wally without mbed
- wally with mbed with acceleration off
- wally with mbed with acceleration on

should also work on esp32s2, esp32c3, esp32c6, esp32p4, etc

Please note:

There's 4 implementations: software, parallel engine, block engine, dma

SOC_SHA_SUPPORT_PARALLEL_ENG is only defined for esp32. esp32s3 uses DMA

SOC_SHA_ENDIANNESS_BE is not available in release branches, only in master and only for esp32 AFAIK

example of how idf checks for things https://github.com/espressif/esp-idf/blob/master/components/mbedtls/port/sha/esp_sha.c#L19

We depend on `#include <soc/soc_caps.h>` to get SOC_SHA_SUPPORT_PARALLEL_ENG (which wasn't available to wally before!!!) and we depend on `#include <sdkconfig.h> to get CONFIG_MBEDTLS_HARDWARE_SHA which is necessary to distinguish between mbedtls with or without hw acceleration.

instead of including sdkconfig.h (and or soc_caps.h) we could have other flags like `HAVE_MBEDTLS_SHA256_H` and `HAVE_MBEDTLS_SHA512_H`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,9 +89,12 @@ $ brew install swig
 - `--enabled-standard-secp`. Excludes support for features that are unavailable in
    the standard [libsecp256k1 library](https://github.com/bitcoin-core/secp256k1).
 - `--enable-mbed-tls`. Use mbed-tls hashing functions if available. This typically
-   results in faster hashing on embedded platforms such as STM32. Note that the user
-   must define `MBEDTLS_SHA256_ALT` and/or `SOC_SHA_SUPPORT_PARALLEL_ENG` matching the
-   SOC support when compiling the library. (default: no)
+   results in faster hashing via hardware on embedded platforms such as ESP32.
+   Note that the caller must ensure that ``sdkconfig.h`` and ``soc/soc_caps.h``
+   are available when compiling, e.g. by setting the `CFLAGS` environment variable
+   before calling configure.
+   must define/undefine `MBEDTLS_SHA256_ALT`, `SOC_SHA_SUPPORT_PARALLEL_ENG`
+   matching the SOC support when compiling libwally. (default: no)
 - `--enable-coverage`. Enables code coverage (default: no) Note that you will
    need [lcov](http://ltp.sourceforge.net/coverage/lcov.php) installed to
    build with this option enabled and generate coverage reports.
--- a/src/internal.c
+++ b/src/internal.c
@@ -201,13 +201,23 @@ static void sha256_midstate(struct sha256_ctx *ctx, struct sha256 *res)
     size_t i;
 
 #ifdef CCAN_CRYPTO_SHA256_USE_MBEDTLS
+#ifndef CONFIG_MBEDTLS_HARDWARE_SHA
+#define SHA_CTX_STATE c.MBEDTLS_PRIVATE(state)
+#else
 #define SHA_CTX_STATE c.state
+#endif
 #else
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -90,8 +90,8 @@ $ brew install swig
    the standard [libsecp256k1 library](https://github.com/bitcoin-core/secp256k1).
 - `--enable-mbed-tls`. Use mbed-tls hashing functions if available. This typically
    results in faster hashing on embedded platforms such as STM32. Note that the user
-   must define `MBEDTLS_SHA256_ALT` and/or `SOC_SHA_SUPPORT_PARALLEL_ENG` matching the
-   SOC support when compiling the library. (default: no)
+   must define/undefine `MBEDTLS_SHA256_ALT`, `SOC_SHA_SUPPORT_PARALLEL_ENG`
+   matching the SOC support when compiling libwally. (default: no)
```
jgriffiths, line 94: This needs to be updated, I'll do that if you can rebase this PR over master.

src/internal.c
```diff
@@ -201,13 +201,23 @@ static void sha256_midstate(struct sha256_ctx *ctx, struct sha256 *res)
...
+#else
 #define SHA_CTX_STATE c.state
+#endif
 #else
 #define SHA_CTX_STATE s
 #endif
 
+#if defined(CCAN_CRYPTO_SHA256_USE_MBEDTLS) && \
+    defined(MBEDTLS_SHA256_ALT) && !defined(SOC_SHA_SUPPORT_PARALLEL_ENG)
+    /* HW: Already big endian */
```
jgriffiths, line 215: I think we should make this:

```
#if defined(CCAN_CRYPTO_SHA256_USE_MBEDTLS) && \
    ((defined(MBEDTLS_SHA256_ALT) && !defined(SOC_SHA_SUPPORT_PARALLEL_ENG)) || defined(SOC_SHA_ENDIANNESS_BE)
```

Which will kick in when the idf sdk is updated in the future, what do you think?

greenaddress: I've updated it to be:

```
#if defined(CCAN_CRYPTO_SHA256_USE_MBEDTLS) && \                                
    defined(MBEDTLS_SHA256_ALT) && (!defined(SOC_SHA_SUPPORT_PARALLEL_ENG) || !defined(SOC_SHA_ENDIANNESS_BE))                                                                                                     
    /* HW: Already big endian */      
```

also updated the other for symmetry:
```
#if defined(CCAN_CRYPTO_SHA256_USE_MBEDTLS) && \                                
    defined(MBEDTLS_SHA256_ALT) && (defined(SOC_SHA_SUPPORT_PARALLEL_ENG) || defined(SOC_SHA_ENDIANNESS_BE))                                                                                                       
    /* HW sha engine doesn't allow to extract the midstate */                   
    ctx.c.mode = ESP_MBEDTLS_SHA256_SOFTWARE;      
```

jgriffiths:
Merged with README tweaks, thanks!

PR #433, opened 2023-11-16
Python linkage enhancements

whitslack:
(continued from #419)

* On systems supporting Libtool (i.e., anything but Windows), skip compiling the "amalgamated" sources for the Python module and simply link in the just-built `libwallycore.a` and `libsecp256k1.a`.

* `setup.py`: support optionally building and linking to libwally-core as a dynamic shared object if `WALLY_ABI_PY_WHEEL_USE_DSO=1` is set. Ignored on Windows, where `setup.py` does not build libwally-core.


whitslack:
Oh, I see this is failing. It still works for Gentoo since my ebuild deletes the whole `if not IS_WINDOWS` block since we do the whole library build before starting the Python build, but it no longer works if the Python module is to be built without first building the library. I'll massage this some more later today.

jgriffiths:
This is probably broken since the latest changes only generate a couple of files and then build with the amalgamation, instead of compiling everything twice.

In terms of a mutually acceptable change, I expect the existing build to remain unchanged (i.e. use the amalgamation) for both win and non-win. You should control the optional building with shared linkage/prebuilt libs entirely with your env var.

whitslack:
As of 813950941b160f94ce7a42cde64bf4b241b07e38, I no longer need to hack `setup.py` at all to achieve a holy-grail build, yet it should still work the way you're accustomed to if no special env vars are set.

setup.py
```diff
@@ -26,11 +26,42 @@
...
+        raise default('Invalid flag value: {}'.format(str))
+    return default
+
+def pkg_default(pkg, if_true, if_false=None):
+    b = str_to_bool(pkg, None, None)
+    return if_true if b == True else if_false if b == False else pkg
+
+PKG_SECP256K1 = pkg_default(
+    os.environ.get('WALLY_ABI_PY_WHEEL_USE_PKG_SECP256K1'), 'libsecp256k1_zkp')
+
```
jgriffiths, line 49: This doesn't need to be so general, less code means the actual build logic is clearer. Something like:

```
def _get_system_secp():
    secp_lib = os.environ.get('WALLY_ABI_PY_WHEEL_USE_PKG_SECP256K1')
    if not secp_lib or secp_lib.lower() in ('0', 'n', 'no', 'f', 'false', 'off'):
        return '' # Disabled
    if secp_lib.lower() in ('1', 'y', 'yes', 't', 'true', 'on'):
        return 'libsecp256k1_zkp' # Use the default system wide secp-zkp
    return secp_lib # Use the caller supplied secp library name


PKG_SECP256K1 = _get_system_secp() if USE_LIB != 'no' else ''
```

Should be sufficent. Then change `PKG_SECP256K1 is/is not None` to `if PKG_SECP256K1` /  `if not PKG_SECP256K1` below.

setup.py
```diff
@@ -26,11 +26,42 @@
...
+if PKG_SECP256K1 is not None:
+    try:
+        import pkgconfig
+    except ImportError:
+        print('Error: You specified WALLY_ABI_PY_WHEEL_USE_PKG_SECP256K1, but you do not have'
+            '\nthe pkgconfig Python module installed.',
```
jgriffiths, line 55: nit: `print('Error: WALLY_ABI_PY_WHEEL_USE_PKG_SECP256K1 requires the pkgconfig Python module.')`

setup.py
```diff
@@ -19,7 +35,12 @@ def call(args, cwd=ABS_PATH):
     call(['git','submodule','sync','--recursive'])
     call(['git','submodule','update','--init','--recursive'])
 
-CONFIGURE_ARGS = ['--disable-shared', '--enable-static', '--with-pic',
+if USE_LIB == 'shared':
+    CONFIGURE_ARGS = ['--enable-shared', '--disable-static']
+else:
+    CONFIGURE_ARGS = ['--disable-shared', '--enable-static', '--with-pic']
```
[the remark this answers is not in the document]
whitslack: Shared objects are always compiled as position-independent code. The only reason the static library needs it specified explicitly is that Libtool doesn't know that the static library is going to end up being linked into a shared library.

setup.py
```diff
@@ -66,9 +90,31 @@ def call(args, cwd=ABS_PATH):
...
-    './src/ccan',
     './src/secp256k1/include',
     ]
+library_dirs = [
+    ]
+libraries = [
+    ]
+sources = [
+    'src/swig_python/swig_python_wrap.c',
+    ]
```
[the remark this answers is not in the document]
whitslack: I formatted them this way so that future additions would only need to insert a new line rather than delete and insert. It's the same reason as for putting a comma after the last item in a list, which is a practice that was already applied to this file before I got to it.

setup.py
whitslack: That's not cleanly extensible to future dependencies that also might have pkg-config files.

Also, you have a bad habit of shortening "secp256k1" to "secp". There are other secp curves, you know.

setup.py
whitslack: That's longer than 80 characters, but I'll format it without a line break as you have requested. Should I also remove the line wrapping from the other error/warning messages?

setup.py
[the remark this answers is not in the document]
jgriffiths: Yes, I understand that, but its harmless to specify in the common flags, no? That make the distinction between shared and static clearer here IMO.

setup.py
jgriffiths: I'm fine with the other errors, and for python lines up to 120 characters are fine, cheers.

setup.py
jgriffiths: No problem, I understand. We are very close to the finish line now though, and I appreciate all the effort you have put in and your responsiveness to review.

jgriffiths:
I'm pulling this into the `build_updates` branch and rebasing it back behind the v1.0 version bump. Feel free to update from that branch so when its merged this will be marked merged too, thanks.

jgriffiths:
@whitslack can you confirm that https://github.com/ElementsProject/libwally-core/pull/432 works for you?

whitslack:
> can you confirm that #432 works for you?

@jgriffiths: 1e6a4b090b8efb1b804a1bb509ca2406cde6dfdf ticks every box on my wish list (except for one, which I hadn't previously mentioned):

- [x] `setup.py` does not invoke Autotools if `src/Makefile` already exists, in which case it's assumed that the caller has built libwally-core in the usual way before starting the Python build. The Python module is an optional add-on, and we don't want the Python build to commandeer (or, for that matter, to ignore) the main library build.
- [x] `setup.py` does not invoke Git if it's going to link with libwally-core as a library and it's going to use a system-installed libsecp256k1-zkp. We don't want to require Git as a build-time dependency, and indeed Gentoo ebuilds run in a network-less sandbox during most build phases, so fetching libsecp256k1 sources during the `src_compile` phase isn't possible even if Git is installed on the build system.
- [x] `setup.py` does not recompile any libwally-core code when instructed to link with libwally-core as a library and such library already exists in the build tree. Especially we do not want duplicated code in the Python native extension library since we'll always have `libwallycore.so` installed whenever we have the Python module installed.

The one gripe I still have, which I am okay to address in a later release, is that the libwally-core library differs depending on whether the Python module is configured to be built. Really, `libwallycore.so` should have no awareness of Python in any configuration, and all necessary Python glue should reside in the Python native extension library. This is especially important when we install the Python module concurrently for multiple versions of Python. For instance, on my system I have the wallycore Python module installed for both Python 3.11 and 3.12, but `/usr/lib64/libwallycore.so.1.0.0` specifies a dynamic link with `libpython3.12.so.1.0`. I don't have a good feeling about that link when the module is loaded under Python 3.11. Really, `libwallycore.so` should not link with any `libpython*.so` at all, as such a leak indicates an abstraction leak / layer violation.

jgriffiths:
@whitslack thanks for testing, and your summary of matters from the gentoo POV.

I am hoping to release 1.0.0 imminently so your final point will not make this release. I am open to addressing it in a future release though, even if in an optional manner as per the `setup.py` changes. Feel free to raise an issue with this info for tracking.

I appreciate your patience and responsiveness in getting these changes merged.

PR #434, opened 2023-11-16
crypto: lower the EC_SIGNATURE_DER_MAX_LOW_R_LEN constant to 70

jgriffiths:
This constant is the size of signatures (excluding the sighash byte) produced by wally when grinding. As wally (and libsecp256k1[-zkp]) always produce low-S signatures, the maximum size is 70 bytes plus the sighash byte for low-R signatures.

This affects the dummy witness stack function, making transaction fee estimation more accurate.


LeoComandini:
utACK 168b676d902f39f1101efcc26514e171cba727dc

PR #436, opened 2023-12-21
v 1.1.0

jgriffiths:
- Allow extraction of partially finalized txs from PSBTs
- Allow fetching the number of items in transaction input witnesses
- Minor code and documentation fixes.


src/address.c
```diff
@@ -26,7 +27,7 @@ int wally_bip32_key_to_address(const struct ext_key *hdkey, uint32_t flags,
         (flags & WALLY_ADDRESS_TYPE_P2SH_P2WPKH && version == WALLY_ADDRESS_VERSION_P2PKH_TESTNET))
         return WALLY_EINVAL;
 
-    if (flags == WALLY_ADDRESS_TYPE_P2PKH) {
+    if (flags & WALLY_ADDRESS_TYPE_P2PKH) {
```
lightyear15, line 30: a few lines above you did the same check as ``flags != WALLY_ADDRESS_TYPE_P2PKH``

lightyear15 (approved):
for what it's worth: LGTM

src/address.c
jgriffiths: Yes, the address type flags are mutually exclusive which is what the initial check is for, hence `==`.

However, we may extend the flags in the future, so all other uses in the function should be `&` so that the rest of the function doesn't need changing if new flags are added (if this happens, just the initial check will be changed, to allow the new flags).

PR #448, opened 2024-03-27
wally_scriptpubkey_get_type: only check 1st byte for OP_RETURN scripts

LeoComandini:
Even though a script might not be created with
`wally_scriptpubkey_op_return_from_bytes` (e.g. `<OP_RETURN>`) or it is not standard (e.g. `<OP_RETURN OP_RETURN>`), if it's starting with `OP_RETURN`, it should still be considered a `OP_RETURN` script.

Bitcoin Core checks [1]
```
    bool IsUnspendable() const
    {
        return (size() > 0 && *begin() == OP_RETURN) || (size() > MAX_SCRIPT_SIZE);
    }
```

rust-bitcoin does [2]
```
    pub fn is_op_return(&self) -> bool {
        match self.0.first() {
            Some(b) => *b == OP_RETURN.to_u8(),
            None => false,
        }
    }
```

----
[1] https://github.com/bitcoin/bitcoin/blob/d04324a7056a735c1127ba8ccdc720a16e7281a3/src/script/script.h#L552-L555

[2] https://github.com/rust-bitcoin/rust-bitcoin/blob/9df59639cec214bd9363d426335923611a304119/bitcoin/src/blockdata/script/borrowed.rs#L350-L354

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## Unreleased
+
+### Fixed
+- wally_scriptpubkey_get_type: mark all scripts starting with OP_RETURN as
+  WALLY_SCRIPT_TYPE_OP_RETURN.
+
 ## Version 1.2.0
 
 ### Added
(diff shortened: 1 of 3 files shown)


jgriffiths:
utack fcd39ceed6d9f3f026ccad7ce163a342bc2c0456, will merge once the CI completes thanks.

PR #452, opened 2024-07-04
psbt: don't serialize witness data for input non-witness utxos

jgriffiths:
Some test vectors in BIP 0174 are incorrect - they include witness data in non-witness UTXOs, which is not output (at least now) by core.

Accept witness data if provided, but do not output it. This matches cores behaviour as can be seen by running utxoupdatepsbt on the test samples that have been fixed here. We keep some old test cases from the BIP to prove wallys re-serialization matches core.


JamieDriver:
I agree from my own testing w/ core that 'p2wpkh' and 'p2sh-p2wpkh' inputs have both `witness_utxo` and `non_witness_utxo` sections, whereas 'p2pkh' inputs only have a `non_witness_utxo` section.

On this basis, utack [1eb727c](https://github.com/ElementsProject/libwally-core/pull/452/commits/1eb727c985b73989553b1ec4998871293a859d9b)

PR #453, opened 2024-07-05
psbt_sign_bip32: use the derived key to sign

jgriffiths:
Fixes signing with parent keys, since the per-input signing call does  not derive keys to sign with.


JamieDriver:
utack [c5e6243](https://github.com/ElementsProject/libwally-core/pull/453/commits/c5e6243c722de77f6978fdb87d57832ecacc1c6c)  

This is similar to the [afaik working!] code in [the Jade repo](https://github.com/Blockstream/Jade/blob/7bdd03103fdee436ff4a1a7bbe6e13202efe73ba/main/process/sign_psbt.c#L748) where the derived key is passed into `wally_psbt_sign_input_bip32()`

PR #462, opened 2024-10-19
Enable and fix for windows build by cmake project file

qwerity:
Enabled by default cmake project files.
Also fixed the Windows build.
Tested for VS2019/VS2022.


jgriffiths:
Hi @qwerity 

Thanks for this! Can I ask you to make the following changes?

- The secp update should be in its own commit. I'm about to update master shortly so I will update secp there and you can rebase this PR from master.
- Please remove the indentation from the `#include` line protected by `#ifndef`
- Please either remove the cmake file renaming or add it as a separate commit. We currently have a downstream fork of wally that uses cmake and are in the process of merging them. Until that is complete (i.e. they can switch to our upstream version here) we are keeping the cmake files renamed so they can continue to take our changes. Once that is done we will rename the files.

jgriffiths (commented):
Hi,

Left some review comments. It appears you are not the original author of this code - is the original author OK with you submitting this PR?

PR #465, opened 2024-11-18
docker: add arm64 support

DeviaVir:
Demo available at https://hub.docker.com/r/blockstream/wallycore/tags


jgriffiths:
@DeviaVir This breaks a few things; I've merged an updated version here https://github.com/ElementsProject/libwally-core/pull/467 which should work for you.

Please LMK if you have any issues after updating to those commits, cheers.

PR #468, opened 2024-11-26
Docker: remove default

DeviaVir:
I noticed that the ARM build was using amd64 as folder, this was fixed after removing the default value here.

### DIFF
--- a/contrib/Dockerfile_bullseye
+++ b/contrib/Dockerfile_bullseye
@@ -7,7 +7,7 @@ FROM debian:bullseye@sha256:01559430c84e6bc864bed554345d1bfbfa94ac108ab68f39915c
 WORKDIR /root
 COPY bullseye_deps.sh ./deps.sh
 COPY requirements.txt ./contrib/requirements.txt
-ARG TARGETARCH=amd64
+ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
 ENV JAVA_HOME=/usr/lib/jvm/java-1.11.0-openjdk-${TARGETARCH}
 RUN ./deps.sh && rm ./deps.sh


jgriffiths:
Hi @DeviaVir 

I'm building `arm64` here with:

```
DOCKER_BUILDKIT=1 docker build . -t greenaddress/wallycore -f Dockerfile_bullseye --platform linux/arm64 --build-arg TARGETARCH=arm64
```

And skipping `--platform` and `--build-arg` for amd64 (which I consider the default build).

My understanding is that you need to pass in or default `TARGETARCH` otherwise e.g. `JAVA_HOME` will not be set to the correct directory in the resulting image. I'd prefer not to have to force anyone building the default image to change their build command.

Can you confirm the above command (or its buildx variant) works for you? If so I will document the arm64 instructions in the dockerfile and leave the default there.

DeviaVir:
> Hi @DeviaVir
> 
> I'm building `arm64` here with:
> 
> ```
> DOCKER_BUILDKIT=1 docker build . -t greenaddress/wallycore -f Dockerfile_bullseye --platform linux/arm64 --build-arg TARGETARCH=arm64
> ```
> 
> And skipping `--platform` and `--build-arg` for amd64 (which I consider the default build).
> 
> My understanding is that you need to pass in or default `TARGETARCH` otherwise e.g. `JAVA_HOME` will not be set to the correct directory in the resulting image. I'd prefer not to have to force anyone building the default image to change their build command.
> 
> Can you confirm the above command (or its buildx variant) works for you? If so I will document the arm64 instructions in the dockerfile and leave the default there.
> 
> edit: documented in this WIP PR: #469

I think you're misunderstanding how this is supposed to work, the idea is that you create a single build for both, so you never have to pass the TARGETARCH, docker does that for you.

This way, whichever architecture you're on, you can pull the same images.

jgriffiths:
> so you never have to pass the TARGETARCH, docker does that for you.

I think this is only true for `buildx`, not `build`?

The former has now replaced the latter in current docker releases, but isn't available on bullseye.

jgriffiths:
Merged a bookworm image which handles multi-arch, while leaving bullseye as defaulting to amd64.

Issue #470, opened 2024-12-11
Unable to use the library with Common js project 


Faisal-3389:
I tried using the library as recommened in my CJS project

`If you're using CommonJS, the module can be loaded asynchronously using const wally = await import('wallycore') or import('wallycore').then(wally => { ... }).
`

but i fails with 
`require() of ES Module /node_modules/wallycore/src/index.js from .../createLBTCDepositAddress.ts not supported.\nInstead change the require of index.js in ./createLBTCDepositAddress.ts to a dynamic import() which is available in all CommonJS modules.`

I tried using esbuild to import the project as cjs module but it is not able to convert due to 
`✘ [ERROR] Top-level await is currently not supported with the "cjs" output format`

Is there a recommended way to use it in the CJS node project, or any example project i can follow. please let me know, thank you!

jgriffiths:
Hi @Faisal-3389 

I'm not sure of your exact setup, but that error seems to be a common error when using CJS. Can you check out the advice in the following links and see if it helps?

- https://www.reddit.com/r/node/comments/1eqgd36/importing_with_import_vs_require/
- https://medium.com/@chengbotier1/solve-the-compatibility-issue-of-typescript-node-tsx-c0b2574636eb

If you are able to resolve the issue please let me know what worked and I will update the documentation...

shesek:
> but i fails with `require() of ES Module /node_modules/wallycore/src/index.js`

This appears to happen because you were using `require()` instead of `import()`.

> `Top-level await is currently not supported with the "cjs" output format`

As the message says, you cannot use `await` in the top-level. Instead, you can use `then()` as mentioned in the README:
```js
import('wallycore').then(wally => {
  console.log(wally.bip39_get_word(null, 10))
})
```

Or alternatively, wrap your code in an IIFE:
```js
(async function() {
  const wally = await import('wallycore')
  console.log(wally.bip39_get_word(null, 10))
})()
```

PR #491, opened 2025-04-02
wasm: Update EMCC_OPTIONS default behavior

shesek:
The existing `build_wasm.sh` script always enables `WASM_BIGINT`, which is necessary for the NPM WASM package but has to be disabled for the esplora build.

This commit changes the default behavior to enable all options necessary for the NPM package (`WASM_BIGINT` plus `EXPORT_ES6` too), while allowing the esplora build to override it with a separate set of options (`MODULARIZE` and `EXPORT_NAME` only).

The [esplora build process](https://github.com/Blockstream/esplora/blob/0732c35c674c7c077399b9f75a9149116bf09f08/Dockerfile.deps#L7) already sets the appropriate `EMCC_OPTIONS`, so no changes necessary there. With this PR merged, we can update the esplora build to use the latest libwally (it currently uses an [older version](https://github.com/ElementsProject/libwally-core/blob/ea984fc07f4f450b33d4eb78756f25f553e60b44/tools/build_wasm.sh) which did not set `WASM_BIGINT`, and would not work against libwally's current master).

This also updates the libwally gitlab CI build to use the correct `EMCC_OPTIONS` for esplora.

However I'm not quite sure why these options are duplicated in this repo. Wouldn't it be better to let the esplora build process control them? So for example, if a new `EXPORTED_FUNCTION` is needed for esplora the changes would be contained in the esplora repo and won't require sending PRs to two separate repos.


shesek:
> With this PR merged, we can update the esplora build to use the latest libwally

I tested esplora against this PR branch (using the Docker build) and it appears to work, however one more minor change is necessary (the destination directory [was changed](https://github.com/ElementsProject/libwally-core/commit/f4229b58c7d21e699c60983f7edb44266651a4e0#diff-099d172147ace15dd52979d57196e96d3c9dc8fa77f92fb9e216e1301600c35fL46-R46) from `wally_dist` to just `dist`). I'll open a PR on esplora once this PR is merged.

jgriffiths:
> However I'm not quite sure why these options are duplicated in this repo. Wouldn't it be better to let the esplora build process control them? 

The esplora build process changes (which are either in-house, in-flight or have landed, I don't follow esplora commits) moved to split-arch, x86 and arm64. In that process, rebuilding the wally wasm was happening with different emcc versions since arm64 release lag quite a bit. Since the resulting wasm should be consistent no matter where esplora is built, it was decided to make esplora wasm release part of the wally release process, so esplora can just download/sha256sum check the build binary and include it directly instead of rebuilding it pointlessly every time.

The use of wally for unblinding in esplora hasn't changed in 4+ years and there are no plans for further changes AFAIU. The only reason its being changed now and the current b0rkage was noticed is the move to supporting arm64. With usable binary releases and no need to rebuild wally, the build in esplora can be greatly simplified, and wally will take responsibility for ensuring the file remains compatible.

jgriffiths:
@shesek Merged, thanks. Would you be able to have a look at https://github.com/ElementsProject/libwally-core/issues/470, I have no idea whats going on in JS land tbh...

PR #498, opened 2025-07-16
Mbed sha256

jgriffiths:
Cleanups for the ccan sha code and a fix for tx_io hashing for non-default sha256 implementations (affecting taproot txs with more than 1 taproot input).


greenaddress:
utack 1117e5bf2df741a8830cea6085dded19a70a6ec3

PR #501, opened 2025-07-31
don't embed Python code into `libwallycore.so`

whitslack:
Embedding SWIG Python code into `libwallycore.so` ties the library to a specific version of Python, thereby precluding concurrent installation of the `wallycore` Python module for multiple implementations of Python on the same system if the Python modules are linked with a system-wide `libwallycore.so`. On Gentoo at least, linking `libwallycore.so` with `libpython3.14.so` and then attempting to use the `wallycore` Python module from Python 3.11, 3.12, or 3.13 causes an immediate segfault.

Rather than injecting Python-specific glue into `libwallycore.so`, we can and should keep it contained within the Python native extension library that we build for each Python implementation. The Python wheel actually already compiles `swig_python_wrap.c` and links it into the native extension library, so linking it into `libwallycore.so` as well is redundant and harmful.

* Remove `libswig_python.la` from `libwallycore_la_LIBADD`, and expunge its existence from `Makefile.am` entirely since it's now unused.

* The Python wheel build for each Python implementation compiles `swig_python_wrap.c` against the headers for that particular Python implementation, ensuring that the resulting native extension library matches the ABI of the Python implementation for which it is installed. This may be a different Python implementation than the one found by Autoconf, the only relevance of which now is in finding the interpreter with which to run the tests.

* Since we no longer link `libwallycore.so` with `libpython*.so`, there should no longer be any manylinux compatibility issues, so drop the `--enable-python-manylinux` Autoconf option and the `PYTHON_MANYLINUX` Automake conditional. This also means that the test programs no longer need to link with `$(PYTHON_LIBS)` since `libwallycore.la` now never has any dependence on Python (implicit or explicit).

* Note that the Python native extension libraries do _not_ explicitly link with `libpython*.so` either. This is correct, as the dynamic linker resolves their undefined `Py*` symbols when it loads them into the Python interpreter.

After applying these changes, the Gentoo ebuild for libwally-core 1.5.0 now successfully runs the SWIG Python tests on all currently supported versions of Python (3.11 through 3.14), all using the same `libwallycore.so.6`.

----

For reference, this is a follow-up to a [comment](https://github.com/ElementsProject/libwally-core/pull/433#issuecomment-1817715361) I made back in November 2023:

> The one gripe I still have, which I am okay to address in a later release, is that the libwally-core library differs depending on whether the Python module is configured to be built. Really, `libwallycore.so` should have no awareness of Python in any configuration, and all necessary Python glue should reside in the Python native extension library. This is especially important when we install the Python module concurrently for multiple versions of Python. For instance, on my system I have the wallycore Python module installed for both Python 3.11 and 3.12, but `/usr/lib64/libwallycore.so.1.0.0` specifies a dynamic link with `libpython3.12.so.1.0`. I don't have a good feeling about that link when the module is loaded under Python 3.11. Really, `libwallycore.so` should not link with any `libpython*.so` at all, as such a link indicates an abstraction leak / layer violation.

That bad feeling has indeed blossomed into an actual segfault as of Python 3.14.

At that same time, @jgriffiths [noted](https://github.com/ElementsProject/libwally-core/pull/433#issuecomment-1817915976):

> I am open to addressing [your final point] in a future release[…]. Feel free to raise an issue with this info for tracking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,6 @@ $ brew install swig
 - `--enable-swig-python`. Enable the [SWIG](http://www.swig.org/) Python
    interface. The resulting shared library can be imported from Python using
    the generated interface file `src/swig_python/wallycore/__init__.py`. (default: no).
-- `--enable-python-manylinux`. Enable [manylinux](https://github.com/pypa/manylinux)
-   support for building [PyPI](https://pypi.org/) compatible python wheels. Using
-   the resulting library in non-python programs requires linking with `libpython.so`.
 - `--enable-swig-java`. Enable the [SWIG](http://www.swig.org/) Java (JNI)
    interface. After building, see `src/swig_java/src/com/blockstream/libwally/Wally.java`
    for the Java interface definition (default: no).
--- a/configure.ac
+++ b/configure.ac
@@ -338,11 +338,10 @@ AC_SUBST([libsecp256k1_LIBS])
 #
 # Python facilities
 #
-AC_ARG_ENABLE(python-manylinux,
-    AS_HELP_STRING([--enable-python-manylinux],[enable manylinux Python compatibility (default: no)]),
-    [python_manylinux=$enableval], [python_manylinux=no])
-AM_CONDITIONAL([PYTHON_MANYLINUX], [test "x$python_manylinux" = "xyes"])
-
+dnl We set the 'optional' flag here because manylinux cibuildwheel containers
+dnl don't include libpython*.so, and thus AX_PYTHON_DEVEL fails its final sanity
+dnl check. We actually don't link against libpython*.so anyway, so all we care
+dnl about is that we get a good value of $PYTHON_CPPFLAGS to pass to SWIG.
 AX_PYTHON_DEVEL([>= '3.9.0'], [true])
 AM_CONDITIONAL([HAVE_PYTHON], [test "x$ax_python_devel_found" = "xyes"])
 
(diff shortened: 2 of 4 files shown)


jgriffiths:
Hi @whitslack the CI does not like these changes, that will need to be fixed. The CI build results must be uploadable to PyPI as complete wheels, i.e. if a new venv is created then `pip install wallycore==<version>` must result in a usable installed wheel with no external dependencies other than libc.

You may need to gate this change behind a configure flag.

jgriffiths:
>  I'm going to have to start carrying a patch alongside the Gentoo ebuild to revert it

If you don't run make check when building your wheels you don't need to revert this surely?

whitslack:
> > I'm going to have to start carrying a patch alongside the Gentoo ebuild to revert it
> 
> If you don't run make check when building your wheels you don't need to revert this surely?

We do run `make check` (at the user's option). Actually, we run `make check-TESTS check-libwallycore` if the user has enabled `FEATURES="test"` and then additionally `make check-swig-java` if they've enabled `USE="java"` and then additionally `make check-swig-python` if they've enabled `USE="python"`.

The reversion turned out not to be a big deal. All I needed was this:
```
use python && sed -e 's:\$(top_builddir)/venv/bin/python \([^-]\):$(PYTHON_TEST) \1:' \
	-e '/\$(top_builddir)\/venv/d' -i src/Makefile.am || die
```
That changes all the (non-PIP) `$(top_builddir)/venv/bin/python` calls back to `$(PYTHON_TEST)` calls and then deletes the two lines that set up the virtualenv and populate it.

jgriffiths:
Good to know, thanks. I'll have a look at the roadmap and make a call next week.

PR #512, opened 2026-01-20
Fix BIP118 ANYPREVOUTANYSCRIPT sighash

instagibbs:
In the current code anytime sh_anyprevout_anyscript is true, so will sh_anyonecanpay will be true. Therefore it will fall through and hash the prevout erroneously. Fix this by checking for APOAS first.

Just tossing this over the wall, fix can be taken in any desired way.


jgriffiths:
@instagibbs Testing this now, will merge with a trivial whitespace fix (tab to spaces). Thanks!

Issue #533, opened 2026-06-29
Expanding Taproot Miniscript and MuSig2 support


odudex:
Hi! 👋

@pythcoiner put together a branch adding Taproot Miniscript and MuSig2 support, and I've been tweaking it. It's now running in [Kern](https://github.com/odudex/Kern/tree/feat/taproot-miniscript), an ESP32P4-based air-gapped signer, where Taproot Miniscript is operational, loading and validating descriptors, signing key-path and script-path.

Branch (based on `release_1.5.4`): https://github.com/odudex/libwally-core/tree/tapscript-and-musig2

We'd really love to upstream this and work together with you to mature it, we think collaborating would make it more robust and reliable for everyone, and we're happy to help in whatever way works best for you.

So the main question: Is this something you'd be open to collaborating on? If so, how would you like us to shape it (e.g. split into smaller PRs, anything else you'd need)?

Thanks for libwally-core! It's been great to build on. 🙏

<img width="400" height="534" alt="Image" src="https://github.com/user-attachments/assets/5fc9ec1c-419d-488c-99ce-2b1f6af4985b" />

<img width="722" height="680" alt="Image" src="https://github.com/user-attachments/assets/15c959aa-1ccf-407a-9d01-20f6a94db633" />

jgriffiths:
Hi @odudex I previously reached out to @pythcoiner to discuss exactly this. We'd love to take the changes!

If you believe your branch is now stable and ready to merge, please rebase and open a pull request to merge it to current wally master. Given the size of the changes the approach I suggested is that I will create a merge branch and add commits from your PR to it as they are reviewed/accepted. Then if needed you can always rebase your branch and we should be able to stay in sync while we work through it. LMK if that sounds OK to you.

jgriffiths:
@odudex Just taking a quick look at your branch, ideally you would take the fixes you have made and merge them back to the commit where the problem they fix was added. In some cases you might need to split up a commit if it has multiple unrelated fixes. If I have to do that while merging we will lose the ability to sync/rebase any other changes I have to make along the way...

odudex:
> ideally you would take the fixes you have made and merge them back to the commit where the problem they fix was added

I thought that would be necessary. I just rebased @pythcoiner 's work to master than added the fixes as follow-ups, but that would be complicated to review indeed.
I'll do a few more rounds of tests and optimizations, then do my best to clean commit history rebasing fixes to appropriate commits, and make a gradual review to a merge branch viable.

[...]

PR #535, opened 2026-06-30
Tapscript & MuSig2 support, with a miniscript decoder and satisfier

odudex:
This PR adds tapscript (taproot script-path) descriptors, a Script→miniscript decoder
and a witness satisfier, MuSig2 (BIP-327) signing, and the descriptor / PSBT plumbing to
use them end-to-end.

The bulk of this work was authored by **@pythcoiner**. My own contributions are some bug fixes,
hardening, the amalgamation build integration folded into the relevant commits and a rebase proposed in #533 .

## Goals

- **Tapscript descriptors** — `tr()` script-path spending: taptree parsing,
  `multi_a`/`sortedmulti_a` fragments, BIP-341 merkle-root/leaf hashing, and the
  tree/leaf/control-block accessor APIs.
- **Miniscript** — a Script→miniscript decoder and a non-malleable, minimum-weight
  witness satisfier (covering tapscript).
- **MuSig2 (BIP-327)** — key aggregation, nonce handling, the full signing pipeline,
  and BIP-328 synthetic-xpub helpers over the secp256k1-zkp musig module.
- **`musig()` descriptors (BIP-390)** — `tr(musig(...))` parsing, address derivation,
  and participant introspection.
- **PSBT** — BIP-371 (taproot) and BIP-373 (PSBT MuSig2) fields, signing, and
  finalization, wired up from descriptors.
- Keep all of the above available in the single-file **amalgamation** build, gated so
  consumers building with `-DBUILD_STANDARD_SECP` (MuSig2 compiled out) still build clean.

## Review structure

The original development history (100+ commits) has been reorganized into **9
self-contained commits**, each a coherent milestone with its own tests. The intent is
to make this reviewable and mergeable **gradually**. Each commit builds, passes its
tests, and can be assessed (or landed) on its own rather than as one monolithic diff.

| # | Commit | What it adds |
|---|--------|--------------|
| 1 | `descriptor: add taproot (tapscript) descriptor support` | `tr()` script paths: `multi_a`/`sortedmulti_a`, taptree parsing & BIP-341 leaf/merkle hashing, script-path address derivation, tree/leaf/control-block accessors |
| 2 | `miniscript: add Script-to-miniscript decoder` | Standalone tokenizer + recursive-descent decoder for all fragments, with stack-overflow / non-minimal-push hardening; BIP-379 differential vectors |
| 3 | `miniscript: add miniscript satisfier` | Per-fragment non-malleable minimum-weight satisfaction/dissatisfaction with overflow-saturated weights and bounded stack usage |
| 4 | `musig2: add MuSig2 (BIP-327) core API` | Key aggregation, nonce gen/agg, signing pipeline, opaque type lifecycle, BIP-328 helpers, bindings; non-aborting handling of malformed cache/session; BIP-327/328 vectors |
| 5 | `descriptor: add musig() key expressions (BIP-390)` | `musig()` as a taproot internal/leaf key, `tr(musig(...))` derivation, participant introspection API; BIP-390 vectors |
| 6 | `psbt: add taproot (BIP-371) and MuSig2 (BIP-373) support` | Taproot + MuSig2 PSBT fields, script-path sighash & signing, finalizers, descriptor-based population; BIP-373 interop tests |
| 7 | `fuzz: add fuzzers for musig() descriptor and PSBT MuSig2 fields` | PSBT MuSig2 fuzzer, `musig()` descriptor fuzz cases + seed corpus |
| 8 | `docs: document the MuSig2 API and add a 2-of-2 PSBT example` | MuSig2 API reference, docs index, CHANGES.md, runnable 2-of-2 MuSig2 PSBT example |
| 9 | `build: add musig and miniscript sources to the amalgamation` | Includes the new sources in the single-file build, gated on `BUILD_STANDARD_SECP` |

## BIPs implemented

BIP-327 (MuSig2), BIP-328 (xpub for aggregate keys), BIP-341 (Taproot), BIP-371
(PSBT Taproot fields), BIP-373 (PSBT MuSig2 fields), BIP-379 (Miniscript),
BIP-390 (`musig()` descriptors).

## Tests
This branch was integrated in [Kern project](https://github.com/odudex/Kern/tree/feat/taproot-miniscript), where tapscript is under tests.

Musig2 support is not currently being tested in a consumer project. I believe @pythcoiner is baking some ideas and can share a north for practical tests.


odudex:
@jgriffiths I changed the PR back to draft after spotting a few issues and opportunities for improvement.

Could you let me know when you're about to start reviewing it? That way I can keep refining the implementation and improving the PR's conciseness and digestibility through small iterations, amending changes into the feature commit as I go.

Once you start the review, I'll stop rewriting the commit(s) you're reviewing and propose any further changes as follow-up commits instead.

jgriffiths:
Hi @odudex now that the PR is composed of a few larger commits, I will be starting with the first commit and reviewing each one essentially independently. This means I will take that commit on its own into a local branch and review/change as needed, without considering the context of future commits (each commit needs to pass independent review and will be updated as needed without considering the potential fallout on following commits).

At the earliest I will start with the first commit at the middle to end of next week. If it makes things easier, I can comment here when I have begun assessing each commit? If you want to merge back a change to a commit later than that one, then you can feel free to do so until the point that I comment here that I have started on it, at which point feel free to push new commits with a comment that it should be merged back to a given commit. For such mergeback commits I will manually make the change(s) needed while reviewing and it can be dropped when the commit in question is merged.

Please let me know if the above approach isn't clear or doesn't work for you. In the meantime until I comment here that I have started, please feel free to merge back/force push changes at will, thanks.

odudex:
> Please let me know if the above approach isn't clear or doesn't work for you. In the meantime until I comment here that I have started, please feel free to merge back/force push changes at will, thanks.

Perfect! I'll keep improving and force-pushing, prioritizing 1st commit, until you start reviewing it. Then I'll wait your feedback and adjust course accordingly.

jgriffiths:
I have started reviewing 2e6bde93c575d0dfa4d5a45571e7f54a89b3e213 (descriptor: add taproot (tapscript) descriptor support)

jgriffiths:
I've made a quick pass through the first commit "descriptor: add taproot (tapscript) descriptor support"- I'll be unavailable for a few days while traveling, so if you are free to look at the changes they are here: https://github.com/ElementsProject/libwally-core/tree/descriptor_updates.

The `MERGEFORWARD` commit is meant to be brought forward to one of the later commits where the constant is referenced (or dropped if it isn't, I haven't checked yet). The `MERGEBACK` commits are meant to be squashed back into the first commit after you've reviewed and are happy with each one. There is commentary on some that you should note - In particular the last one, where I have just noted the changes that I think need to be made, but haven't had time to do yet.

If you are happy with the changes as-is please feel free to squash into this branch and rebase (I hope it isn't too painful - I've tried to avoid gratuitous changes like formatting). If you want to make further changes based on the review so far please feel free, just let me know when I can continue review on the initial commit again.

Other thoughts beyond whats noted in the commits so far:

- I think its worth holding the max depth in the parsed descriptor instead of recursing to compute it. That would allow `path_buf` allocations to use the max depth instead of always allocating 128 * 32 bytes, and allow `wally_descriptor_get_taproot_control_block_len()` to be added with a trivially cheap implementation. Actually with that done we could avoid these allocations altogether and generate the path directly into the output buffer in  `wally_descriptor_get_taproot_control_block()`.
- We need to verify that passing a depth in the flags when parsing correctly limits recursion to that depth (Jade uses this to prevent stack overflow)

Happy to answer any questions on the above, thanks.

odudex:
> I've made a quick pass through the first commit "descriptor: add taproot (tapscript) descriptor support"- I'll be unavailable for a few days while traveling, so if you are free to look at the changes they are here: https://github.com/ElementsProject/libwally-core/tree/descriptor_updates.

Thanks Jon,

I'll go through each commit and its commentary, squash the MERGEBACK commits into the descriptor commit as I review them, and place the MERGEFORWARD constant where it's referenced (or drop it if unused). I'll also work on the changes you noted but didn't have time to make.

On your other points: storing the max depth in the parsed descriptor makes sense. I'll look into it, including generating the path directly into the output buffer. I'll also verify that the depth flag properly limits recursion when parsing.

If questions come up along the way, I'll note them down so we can discuss when you're back.

Safe travels!

odudex:
Thanks for the thorough pass.
All 22 MERGEBACKs made sense to me, so I took them as-is and squashed them into the first commit (added you as co-author). The MERGEFORWARD constant belongs to the later "descriptor: add musig() key expressions" commit, so it moved forward into that one. The branch is rebased, each commit builds and passes the suites individually (elements and no-elements), so review can continue whenever you're back.

A few notes on the open items:

`get_key_xonly_public_key` removal. Accepted for the public API. The PSBT taproot commit needs derived x-only keys internally (populating `TAP_BIP32_DERIVATION` requires each leaf key's derived pubkey as the keypath map key, which the string-based `get_key()` can't provide), so the implementation was kept as an internal `descriptor_get_key_xonly_public_key()` in `descriptor.c`, declared in `descriptor_int.h` and introduced by the PSBT commit itself. The first commit stays exactly as you left it. Your `get_bip32_key()` design looks like the right public interface. I left it as a follow-up rather than growing this PR (the internal helper carries a `FIXME: replace with a public get_bip32_key()` marker so the intent is visible in-tree), unless you'd prefer it done now.

Max taptree depth / control block. Implemented, taking your "avoid these allocations altogether" note to its conclusion: `wally_descriptor_get_taproot_control_block()` now computes the exact control block size up front from the target leaf's depth (a tree walk with no hashing or key generation) and generates the merkle path directly into the output buffer, so the fixed 128×32 `path_buf` allocation is gone entirely, which also removed the need to store a max depth in the parsed descriptor. `wally_descriptor_get_taproot_control_block_len()` is added with the trivially cheap implementation you wanted, wired through the bindings like `get_taproot_leaf_script_len` (this also let `control_block` come off the WASM exclusion list), with C test coverage for exact sizes and out-of-range leaves.

Depth-in-flags recursion limit. Already enforced: `wally_descriptor_parse()` pre-scans the source with `get_max_depth()`, which counts `{`/`}` as well as `(`/`)`, so taptree nesting counts toward the caller-supplied `WALLY_MINISCRIPT_DEPTH_MASK` limit and over-deep descriptors are rejected before any recursive parsing runs. Independently, `parse_taptree_impl()` enforces the unconditional BIP-341 cap of 128. There's an existing test covering the flags path ("Taptree branches count towards the caller-supplied maximum depth" in `test_descriptor.c`).

`const.js` discrepancy. While resolving the MERGEFORWARD I found the wasm mirror had `WALLY_MS_IS_MUSIG = 0x1000`, colliding with `WALLY_MS_IS_TAPSCRIPT` (the header value is `0x2000`). Fixed in the `musig()` commit.

Ripple effects in later commits. Two things you'll see when you get to them, both deliberate: the musig core commit's regenerated bindings had re-introduced the `get_key_xonly_public_key` mirrors and the old (variant-arg) taproot getter prototypes, those were scrubbed during the rebase, and the PSBT commit's call sites were updated to the new variant-less getter arities and to `bip341_tapleaf_hash`/`get_taproot_leaf_script_len`.

Beyond the above, no other changes were made to the first commit. It's ready for you to resume review when you return.

I tested basic tapscript functionality from this branch in Kern, and will do more tests next week.

jgriffiths:
Awesome stuff @odudex - I'll sync the changes and continue review while traveling, thanks!

jgriffiths:
Hi @odudex I've made pass #2 through the first commit. Like before, the updates are here https://github.com/ElementsProject/libwally-core/tree/descriptor_updates if you could take a look and squash back if you are happy.

I will make a pass #3 once this set of changes is squashed, I think that will likely be the final set of changes before merging. Again please LMK if you have any issues/questions, thanks.

odudex:
Hi @jgriffiths, all 17 commits squashed back, no issues. I only did an update to `src/wasm_package/src/const.js` with proper values for `WALLY_MS_IS_TAPSCRIPT` and `WALLY_MS_IS_TAPROOT`

jgriffiths:
Hi @odudex I've made pass number 3 through the first commit. Like before, the updates are here https://github.com/ElementsProject/libwally-core/tree/descriptor_updates if you could take a look. Note there are some bugfixes in this round (depth check off-by-one and fixed tr() embedded in ct()).

I think I will add the descriptor bip32 key derivation call on top of master and then rebase the first commit on top of it, so I can then remove the function returning the internal key. I wont start that rebase until you've reviewed and are happy/have squashed , cheers.

odudex:
> Hi @odudex I've made pass number 3 through the first commit.

Hi @jgriffiths, thanks for the third pass.
I reviewed all 14 commits and squashed them into the first commit. The ct() fixes were great catches. Two deviations from your branch:

**Depth check:** I kept your refactor but reverted the check to `depth > WALLY_DESCRIPTOR_TAPTREE_MAX_DEPTH`. A leaf at depth 128 has a merkle path of exactly m=128 hashes, which is the BIP-341 maximum (control block of 33 + 32*128 bytes), so it's valid. Core agrees: its parser errors only when `branches.size() > TAPROOT_CONTROL_MAX_NODE_COUNT` after pushing, i.e. it accepts 128 nesting levels and rejects 129. With `>=` we'd reject depth-128 descriptors that core accepts. I added a `check_taptree_depth_limit()` test that builds 128- and 129-deep trees and asserts the boundary (128 → OK, 129 → EINVAL).
- core parser check (after push, so 128 braces pass): https://github.com/bitcoin/bitcoin/blob/v31.1/src/script/descriptor.cpp#L2481-L2488
- `TAPROOT_CONTROL_MAX_NODE_COUNT = 128`: https://github.com/bitcoin/bitcoin/blob/v31.1/src/script/interpreter.h#L245
- rust-miniscript agrees: a 128-deep tr() parses, 129 fails with "maximum Taproot tree depth (128) exceeded".

**x-only feature test:** in the new test, `wally_descriptor_get_key_features()` was called with the leaf index `i` instead of the `key_index` returned by `get_taproot_leaf_key_index()`. Fixed to use `key_index`.

Everything else is in as-is (plus a few of comment typo fixes). All green here.

Sounds good regarding the bip32 derivation call on master and rebasing the first commit on top of it. I'll keep an eye on it and rebase if needed.

jgriffiths:
Awesome stuf @odudex, great catches on the depth and test indexing var, thanks! I'll update to your branch and start on the key derivation api.

jgriffiths:
Hi @odudex I've implemented `wally_descriptor_derive_bip32_key` on top of (rebased) master, rebased the first commit on that (no conflicts), and added another set of MERGEBACK commits for you to review. As before, the updates are here https://github.com/ElementsProject/libwally-core/tree/descriptor_updates

If you could review the derive_bip32_key changes, that would be great. You should now be able to use it along with `wally_descriptor_get_key_features` to fetch/use the internal key in the later commits. Note that the taproot internal key is always at key index 0 (even in ct() descriptors - there is no need to check for ct() in callers in this case).

odudex:
Thanks @jgriffiths, `derive_bip32_key` is a much better fit than the accessor it replaces. All reviewed and integrated.

### Review of your two commits

One line amended directly into your `derive_bip32_key` commit: `descriptor_get_key`'s `const ms_node *node` is unused in non-Elements builds, so the commit fails `-Werror` on the `--enable-elements=no` CI job. Your `reduce elements ifdef-iness` MERGEBACK fixes it later, so only that commit was affected. I moved the declaration inside the `#ifdef` so every commit builds clean.

The rest is in the separate `crypto/descriptor: review fixes` commit, so you can merge, amend or drop it as one unit:

- **`node_derive_key` doesn't mark public-only results as public.** `output` is cleared on entry, leaving `priv_key[0] == BIP32_FLAG_KEY_PRIVATE`, so `bip32_key_is_private()` returns true for a bare public key holding an all-zero private key. Now set to `BIP32_FLAG_KEY_PUBLIC` in the three public-only paths.
- **`ec_public_key_compress`/`_decompress` skip validation on the no-op path.** The memcpy shortcut returns `WALLY_OK` for an invalid key whenever no format conversion is needed, e.g. `compress()` on 33 bytes of garbage. They now always parse, which also collapses the duplicated argument checks into `pk_convert_impl` (renamed from `pk_compress_impl` since it serializes either way) and clears the parsed pubkey like its siblings in sign.c.

While squashing the MERGEBACKs into our first commit: stray leading space on `wally_descriptor_get_taproot_leaf_num_keys`, `num_keys_impl` -> `leaf_keys_impl`, `key_index` -> `key_position` for header parity, and the C99 mid-block declarations in the ctest hoisted to match the surrounding style.

One open question: `wally_descriptor_derive_bip32_key` takes `uint32_t index` while `wally_descriptor_get_key` / `_get_key_features` take `size_t index`.

### Adopting derive_bip32_key downstream

`node_derive_key` needed a `musig()` branch, since a `musig()` aggregate is itself key index 0 and would otherwise return `WALLY_ERROR`. `generate_musig` is refactored into a `musig_derive_key()` filling an `ext_key`, used by both paths. psbt.c gained a small static `descriptor_get_taproot_internal_key()` over `derive_bip32_key` for the two removed-API call sites.

The remaining commits also picked up some fixes of our own, found while running different CI configurations.

jgriffiths:
Hi @odudex Apologies for the delay, we are somewhat overloaded at the moment, I hope to get back to this in a few days. Master has been updated if you wouldn't mind rebasing in the meantime, thanks!

odudex:
> Hi @odudex Apologies for the delay, we are somewhat overloaded at the moment, I hope to get back to this in a few days. Master has been updated if you wouldn't mind rebasing in the meantime, thanks!

Hi @jgriffiths, no problem! It was a quite unusual week. There's no reason to rush on my side.
PR branch rebased!

jgriffiths:
Hi @odudex -  https://github.com/ElementsProject/libwally-core/tree/descriptor_updates is updated to your branch plus a couple of final fixes:

**derive_bip32_key**:

> node_derive_key doesn't mark public-only results as public

Added a mergeback commit to use `bip32_key_strip_private_key` to mark the keys as public instead of setting the values ourselves. Also to wipe the output key on entry.

> One open question:

Migrated to size_t for consistency, good spotting, thanks.

**add taproot (tapscript) descriptor support**:

> key_index -> key_position for header parity

Added a mergeback commit to reapply this and fix the header/generated wrappers. wally uses _index for indices, not _position.

Renamed the impl again also, but chose a name that won't clash in the future with amalgamated builds.

Assuming this round passes review without changes I will merge these two commits, thanks.

odudex:
Hi @jgriffiths!
Fixes and improvements were integrated.

jgriffiths:
Hi @odudex - Merged. Moving on to commit number 2!

odudex:
> Hi @odudex - Merged. Moving on to commit number 2!

Great news @jgriffiths @pythcoiner !
Branch rebased!

odudex:
Hi @jgriffiths ,
I rebased this PR with latest commit from master, and also did another review round with multiple fixes in `miniscript: add Script-to-miniscript decoder` to increase consistency with core. I hope I didn't got you in the middle of the review.

jgriffiths:
> I hope I didn't got you in the middle of the review.

You did! If you are going to change commits please do so via MERGEBACK commit which I can ack and then you can squash, cheers.

I started making some smaller MERGEBACK tweaks while reviewing the second commit, but after a while it became clear that the overall structure needs reworking I think.

The intention of the second commit is to be able to convert script into a parsed miniscript AST. it does this by tokenizing the script, then iterating it backwards to construct a copy of the parsed tree directly (albeit containing extra decoder node types).

Issues I have with this:

The tokenizer iterates the script and allocates tokens, which doesn't seems to be needed at all. A forward iterator interface is sufficient to tokenize the script; this doesn't require any extra memory.

Other than allocating a token stack, each token contains an unneeded copy of any pushed data while the underlying script bytes are valid for the whole tokenization process.

The tokenizer itself has checking logic split between tokenization and use which breaks layering.

Some other strange stuff like it rejects non-miniscript push sizes but then parses push opcodes that it will never allow the data push for (PUSH2/PUSH4), and doesn't enforce minimal pushes in these cases.

Once tokenized, the list of tokens is parsed backwards to reconstruct the AST. This makes the construction code hard to reason about. But directly constructing the AST also means the same miniscript expression now has a different AST representation depending on whether it was made from miniscript descriptor text or a script. That seems undesirable for a few reasons.

If the AST code instead converted script to miniscript, it could be parsed with the existing descriptor parsing code, leaving a single representation. But more importantly, this process can be round-tripped with the existing tests:  miniscript -> script -> miniscript roundtripping proves both side of the conversion and ensures that the AST representation for both will always be the same.

jgriffiths:
I have started a branch with a general non-allocating script tokenizer which I will push for comment soon. Also a fix to the existing scriptint parsing which lets the duplicate logic in the second commit be removed.

odudex:
> You did!

Oof I'm sorry! I did so many changes it is almost a 3rd front. Next time I'll use the MERGEBACK approach.

I'm glad you have ideas for, and is already cooking a more efficient structure with the non-allocating script tokenizer. Now is the best time to evaluate all possible architectures.
Let me know if I can help with something, or when it's ready to review. I'll ensure subsequent commits to be compatible with eventual changes.

